PR #6, opened 2018-06-20
refactor and move forms to backend, introduce application context

LCairns:


zuzak:
Session secret should be configurable?

PR #30, opened 2018-07-09
Add bulk upload form

zuzak:
Doesn't prompt user for the case type at the moment, alas.


zuzak:
Requires UKHomeOffice/hocs-workflow#5 upstream

zuzak:
superseded by #35

PR #36, opened 2018-07-10
Add backlink; add concept of secondary action

zuzak:
The submit button on the form is currently hardcoded to exist as the
very bottom form element. This is problematic as usually we want the
primary action button to be aligned to the left edge of the form, and in
the user's line of sight. At the moment, secondary actions come
beforehand which is confusing.

This commit adds the concept of secondaryActions separate to the main
form to allow for buttons and other elements after the main submit
button. We whitelist elements to fairly innocuous ones to ensure that
the primary button remains primary.

This commit also adds a back link component to go in this
secondaryActions thing because "cancel" shouldn't
be a similar precedence than "next step".

### DIFF
--- a/server/forms/bulk-create.json
+++ b/server/forms/bulk-create.json
@@ -1,0 +1,48 @@
+{
+  "form": {
+    "schema": {
+      "action": "bulk_upload",
+      "callback": {
+        "action": "case",
+        "value": "documents"
+      },
+      "title": "Create cases in bulk",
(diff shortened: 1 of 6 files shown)


zuzak:
Not particularly bothered if we fix this like this, or if we just stop hardcoding the submit button.

Either way, I don't like how "cancel" is currently a blue button instead of an unobtrusive link -- it'd be nice if we fixed that at least.

This PR needs closing before #35 probably

PR #43, opened 2018-07-13
Unify Normal and Bulk create

nickmcmahonHOD:


src/shared/common/forms/composite/bulk-document-add.jsx
```diff
@@ -0,0 +1,69 @@
+import React, { Component, Fragment } from 'react';
```
zuzak: Try reusing `document-add.jsx` and simply changing the properties in the `.json` file to overwrite the bits you want to change.

server/config.js
```diff
@@ -4,7 +4,7 @@ const workflowAuth = (process.env.WORKFLOW_BASIC_AUTH || 'UNSET:UNSET').split(':
 const config = {
     applications: {
         render: {
-            clientSide: process.env.USE_CLIENTSIDE || true,
+            clientSide: process.env.USE_CLIENTSIDE || false,
```
zuzak: If not accidental, this should be in a separate pull request.

PR #49, opened 2018-07-17
Display footer and phase banner by default

zuzak:
The crest makes me feel important


nickmcmahonHOD:
no because tech lead.

PR #74, opened 2018-08-02
Stop storing X-Auth-Token

zuzak:
We now check the presence of X-Auth-Token, but there's no need to do
anything with it -- the other X-Auth-* headers are sufficient, so this
commit removes the bit where we store the token.

In the wrong hands you can use the contents of X-Auth-Token to
impersonate a user and their roles; so let's keep our hands clean of it
as much as possible.


zuzak:
I'm mean, we're essentially sending the user's password with each request 🤔

nickmcmahonHOD:
.

PR #115, opened 2018-10-15
Lc workstack refactor

LCairns:


server/services/list.js
```diff
@@ -122,8 +65,33 @@ const lists = {
             'X-Auth-UserId': user.id,
             'X-Auth-Roles': user.roles.join()
         });
-        const dashboadData = response.data.activeStages;
-        return temporaryWorkstackToDashboardAdapter(dashboadData);
+        const workstackData = response.data.activeStages
+            .sort((first, second) => first.caseReference > second.caseReference ? 1 : -1);
```
zuzak: True is bigger than false, so you might not need this ternary operator
```js
true > false → true
true < false → false
```

PR #266, opened 2019-04-05
fix document case note formatting

SWhitlam:


nickmcmahon01:
This is quality work.

PR #289, opened 2019-05-10
Add tests to build

SWhitlam:


Dockerfile
```diff
@@ -17,6 +17,7 @@ RUN groupadd -r ${GROUP} && \
 WORKDIR /tmp
 COPY . /tmp
 RUN npm --loglevel warn install --development --no-optional
+RUN npm test
```
nickmcmahon01: Is this the best way, can't we add it to the drone?

PR #305, opened 2019-07-04
Making sure that keycloak version is consistent (5.0.0)

hubert-raich:
Updating port for workflow service running locally

### DIFF
--- a/server/config.js
+++ b/server/config.js
@@ -30,7 +30,7 @@ const config = {
             }
         },
         server: {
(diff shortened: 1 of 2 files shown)


server/config.js
```diff
@@ -30,7 +30,7 @@ const config = {
             }
         },
         server: {
-            WORKFLOW_SERVICE: process.env.WORKFLOW_SERVICE || 'http://localhost:8081',
+            WORKFLOW_SERVICE: process.env.WORKFLOW_SERVICE || 'http://localhost:8091',
```
nickmcmahon01: Is there a reason for changing the default port, does something else run on 8081?

PR #306, opened 2019-07-19
Changes to 'workflow' docker yml following onboarding experience

MarkWrightBjss:
Ports 8080:8081 are often already in use, so switched.
Default schema to 'workflow' rather than 'public'.
Set the flyway schema name.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -144,7 +144,7 @@ services:
     image: quay.io/ukhomeofficedigital/hocs-workflow
     # image: hocs-workflow-test
     ports:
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -144,7 +144,7 @@ services:
     image: quay.io/ukhomeofficedigital/hocs-workflow
     # image: hocs-workflow-test
     ports:
-    - 8081:8080
+    - 8091:8090
```
hubert-raich: port change incorrect, it should still point to 8080 on docker so please change this to:
- 8091:8080

PR #332, opened 2019-10-25
HOCS-1076-Search for a compensation case

hubert-raich:


server/routes/search.js
```diff
@@ -14,7 +14,10 @@ router.all(['/search', '/api/search', '/api/form/search'], getForm(form, { submi
 router.get('/api/form/search', (req, res) => res.json(req.form));
 
 router.post(['/search/results', '/api/search/results'],
-    processRequestBody(form().getFields()),
+    async (req, res, next) => {
```
lewismuirhead: this could be abstracted into a middleware component

server/routes/search.js
```diff
@@ -33,7 +36,10 @@ router.post(['/search/results', '/api/search/results'],
                 correspondentName: formData['correspondent'],
                 topic: formData['topic'],
                 data: {
-                    POTeamUUID: formData['signOffMinister']
+                    POTeamUUID: formData['signOffMinister'],
```
lewismuirhead: mainly existing code, but I think this handler should definitely be in a middleware with some tests

server/services/action.js
```diff
@@ -86,15 +86,16 @@ const actions = {
...
+                        const documentTags = (await listServiceInstance.fetch('S_SYSTEM_CONFIGURATION')).documentLabels;
+                        response = await createCase('/case', { caseType: context, form }, documentTags[0], headers);
                         clientResponse = { summary: 'Created a new case ', link: `${response.data.reference}` };
                         return handleActionSuccess(clientResponse, workflow, form);
                     }
                     case actionTypes.BULK_CREATE_CASE: {
                         const listServiceInstance = listService.getInstance(uuid(), null);
-                        const documentTags = await listServiceInstance.fetch('S_DOCUMENT_TAGS');
-                        response = await createCase('/case/bulk', { caseType: context, form }, documentTags[0].value, headers);
+                        const documentTags = (await listServiceInstance.fetch('S_SYSTEM_CONFIGURATION')).documentLabels;
```
lewismuirhead: I think: const { documentLabels: documentTags } = await listServiceInstance....
would remove the need for the brackets

server/routes/search.js
hubert-raich: resolved with searchHandler component

server/routes/search.js
hubert-raich: resolved with searchHandler component

server/services/action.js
hubert-raich: done

PR #333, opened 2019-10-29
scrollIntoView is now a function

Raysonviswas:


src/shared/common/forms/__tests__/error-summary.spec.js
```diff
@@ -34,5 +37,10 @@ describe('Form text component', () => {
             )
         ).toMatchSnapshot();
     });
+
+    it('expect createRef to be called', () => {
+        render(<ErrorSummary description="Displaying a list of the errors on the page" />);
+        expect(refSpy).toBeCalled();
```
lewismuirhead: I think we actually need to assert that the scrollIntoView function is called

PR #334, opened 2019-10-31
Change registration form fields to match mock. Add checkboxes.

MarkWrightBjss:
Added checkboxes to dynamic form. Some reformatting to make style police happy. Merge to WCS.

### DIFF
--- a/src/shared/common/forms/checkbox-group.jsx
+++ b/src/shared/common/forms/checkbox-group.jsx
@@ -5,24 +5,29 @@ class Checkbox extends Component {
 
     constructor(props) {
         super(props);
-        this.state = { value: this.props.value };
+        this.state = { value: this.props.value.split(',').filter(cb => cb !== '') };
     }
 
+    stateString() {
+        var value = this.state.value.join();
+        return value;
+    }
 
     componentDidMount() {
-        this.props.updateState({ [this.props.name]: this.state.value });
+        this.props.updateState({ [this.props.name]: this.stateString() });
     }
 
     handleChange(e) {
-        let targetValue = e.target.value + ',';
-        this.setState((currentState) => {
-            if (currentState.value.includes(targetValue)) {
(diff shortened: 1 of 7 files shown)


src/shared/common/forms/checkbox-group.jsx
```diff
@@ -5,24 +5,29 @@ class Checkbox extends Component {
...
-            if (currentState.value.includes(targetValue)) {
-                return { value: currentState.value.replace(targetValue, '') };
-            } else {
-                return { value: currentState.value + targetValue };
-            }
+        var selection = this.state.value;
+        if (selection.includes(e.target.value)) {
+            selection = selection.filter(cb => cb !== e.target.value);
+        } else {
+            selection.push(e.target.value);
```
lewismuirhead: I think you should use array spread to create a new array here

src/shared/common/forms/checkbox-group.jsx
```diff
@@ -5,24 +5,29 @@ class Checkbox extends Component {
...
 
     handleChange(e) {
-        let targetValue = e.target.value + ',';
-        this.setState((currentState) => {
-            if (currentState.value.includes(targetValue)) {
-                return { value: currentState.value.replace(targetValue, '') };
-            } else {
-                return { value: currentState.value + targetValue };
-            }
+        var selection = this.state.value;
```
lewismuirhead: as the setstate depends on current state, use the current state argument from the updater function

src/shared/common/forms/checkbox-group.jsx
MarkWrightBjss: Implemented recommendation.

src/shared/common/forms/checkbox-group.jsx
MarkWrightBjss: Implemented as suggested.

PR #339, opened 2019-11-07
Hocs 1111 according sections appear in random order

Raysonviswas:


server/lists/adapters/case-view.js
```diff
@@ -19,21 +19,19 @@ module.exports = async (template, { fromStaticList }) => {
         .withNoPrimaryAction();
 
     const data = {};
-
     const sections = [];
+    const sortedSections = [];
 
-    await Promise.all(Object.entries(template.schema.fields).map(async ([stageId, fields]) => {
+    await Promise.all(Object.entries(template.schema.fields).sort((a,b) => a[0] > b[0]).map(async ([stageId, fields], index) => {
```
lewismuirhead: as discussed I think this could replace the sort by index below, but also would need to be a standard character sort function a > b return 1, a < b return -1 etc...

PR #340, opened 2019-11-07
HOCS-1260 - Back button within mark up stage

hubert-raich:


server/middleware/stage.js
```diff
@@ -58,9 +58,25 @@ async function allocateCaseToTeamMember(req, res, next) {
...
+async function moveToPreviousStage(req, res, next) {
+    const logger = getLogger(req.request);
+    const { caseId, stageId } = req.params;
+    const { user } = req;
+    try {
+        await workflowService.post(`/case/${caseId}/stage/${stageId}/back`, {
+            userUUID: user.uuid,
+        }, { headers: User.createHeaders(user) });
+    } catch (error) {
+        logger.error(error);
```
lewismuirhead: probably should call next(error) here to pass it to the error handlers

src/shared/common/forms/back-button.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+    handleClick(e) {
+        e.preventDefault();
+        const { dispatch, history } = this.props;
+        const endpoint = `/api/form/case/${this.props.caseId}/stage/${this.props.stageId}/back`;
+        return dispatch(updateApiStatus(status.MOVE_BACK_REQUEST))
+            .then(() => {
+                axios.get(endpoint)
+                    .then(response => {
+                        dispatch(updateApiStatus(status.MOVE_BACK_SUCCESS))
```
lewismuirhead: is this dispatch too early?

src/shared/common/forms/back-button.jsx
```diff
@@ -0,0 +1,89 @@
...
+                            .then(() => {
+                                if (response.data.errors) {
+                                    dispatch(updateApiStatus(status.MOVE_BACK_FAILURE));
+                                } else {
+                                    history.push(response.data.redirect);
+                                }
+                            })
+                            .then(() => dispatch(clearApiStatus()))
+                            .catch(() => {
```
lewismuirhead: I think this would just catch the dispatch failing, which is unlikely, and would be just that the apiStatus was not set.

src/shared/common/forms/back-button.jsx
```diff
@@ -0,0 +1,89 @@
...
+        e.preventDefault();
+        const { dispatch, history } = this.props;
+        const endpoint = `/api/form/case/${this.props.caseId}/stage/${this.props.stageId}/back`;
+        return dispatch(updateApiStatus(status.MOVE_BACK_REQUEST))
+            .then(() => {
+                axios.get(endpoint)
+                    .then(response => {
+                        dispatch(updateApiStatus(status.MOVE_BACK_SUCCESS))
+                            .then(() => {
+                                if (response.data.errors) {
```
lewismuirhead: I think this test should be before setting the success status, which might simplify it a bit

src/shared/common/forms/form-repository.jsx
```diff
@@ -49,69 +50,72 @@ function renderFormComponent(Component, options) {
...
-    case 'button':
-        return renderFormComponent(Button, { key, config });
-    default:
-        return null;
+        case 'backlink':
+            return renderFormComponent(BackLink, { key, config });
+        case 'button':
+            return renderFormComponent(Button, { key, config });
+        case 'backButton':
+            config.action = `/case/${page.caseId}/stage/${page.stageId}/back`;
```
lewismuirhead: this could be inside

src/shared/common/forms/back-button.jsx
```diff
@@ -21,24 +21,17 @@ export class BackButton extends Component {
...
-                            .catch(() => {
-                                dispatch(updateApiStatus(status.MOVE_BACK_FAILURE))
-                                    .then(dispatch(setError({ status: 500 })));
-                            });
-                    })
-                    .catch(error => {
-                        dispatch(updateApiStatus(status.MOVE_BACK_FAILURE))
-                            .then(() => dispatch(setError(error.response)));
-                    });
+                        if (response.data.errors) {
```
lewismuirhead: 👍

PR #343, opened 2019-11-12
HOCS-1250 : add "Previous HOCS reference"

MarkWrightBjss:
New validation option added, "alphanumeric". Merge to WCS.


MarkWrightBjss:
Also added coverage for the error message.

PR #348, opened 2019-11-15
HOCS-1245 : numeric validation

MarkWrightBjss:
Numeric validation. Merge to WCS.

### DIFF
--- a/server/middleware/__tests__/validation.spec.js
+++ b/server/middleware/__tests__/validation.spec.js
@@ -24,6 +24,21 @@ describe('Validators', () => {
         });
     });
 
+    describe('Numeric validator', () => {
+        it('should reject symbols', () => {
+            expect(validators.numeric({ value: '!@£$' })).not.toEqual(null);
+        });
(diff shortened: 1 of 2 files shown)


server/middleware/__tests__/validation.spec.js
```diff
@@ -24,6 +24,21 @@ describe('Validators', () => {
...
+        it('should reject alpha', () => {
+            expect(validators.numeric({ value: 'data' })).not.toEqual(null);
+        });
+        it('should reject alphanumeric', () => {
+            expect(validators.numeric({ value: 'a1b2' })).not.toEqual(null);
+        });
+        it('should accept numeric only', () => {
+            expect(validators.numeric({ value: '1234' })).toEqual(null);
+        });
+    });
```
lewismuirhead: you could also add tests to cover the message/label logic

MarkWrightBjss: Additional coverage added as suggested. I'll do the same to the alphanumeric validator, too.

PR #385, opened 2019-12-12
Hocs 1170 skip case details page

lewismuirhead:


server/middleware/skipCaseTypePage.js
```diff
@@ -0,0 +1,41 @@
...
+        }
+    } else {
+        next();
+    }
+}
+
+async function skipCaseTypePageOnReload(req, res, next) {
+    const workflow = req.params.workflow;
+    const action = req.params.action;
+    if (workflow && action ) {
```
lewismuirhead: Need to check workflow and action parameters here

server/middleware/skipCaseTypePage.js
```diff
@@ -0,0 +1,41 @@
...
+            }
+        } catch (e) {
+            next(e);
+        }
+    } else {
+        next();
+    }
+}
+
+async function skipCaseTypePageOnReload(req, res, next) {
```
lewismuirhead: perhaps rename these two api / non api

PR #404, opened 2020-01-14
HOCS-1402 : calculate total amounts, update totals between form submits

MarkWrightBjss:
Calculate total Awarded and Paid amounts. Merge to WCS.

### DIFF
--- a/src/shared/common/forms/text.jsx
+++ b/src/shared/common/forms/text.jsx
@@ -5,15 +5,13 @@ class Text extends Component {
 
     constructor(props) {
         super(props);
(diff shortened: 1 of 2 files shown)


src/shared/common/forms/text.jsx
```diff
@@ -5,15 +5,13 @@ class Text extends Component {
 
     constructor(props) {
```
lewismuirhead: the whole constructor can go

MarkWrightBjss: Who knew! The whole constructor has gone.

PR #417, opened 2020-01-29
HOCS-1437 numeric text box size

lewismuirhead:


src/shared/common/forms/text.jsx
```diff
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
```
MarkWrightBjss: Never knew this was a thing.

PR #418, opened 2020-01-29
HOCS-1456 fix the display of nested component validation and preserve…

lewismuirhead:
… open state correctly on re-renders

### DIFF
--- a/server/tenantConfig.js
+++ b/server/tenantConfig.js
@@ -38,7 +38,7 @@ async function layoutConfig() {
         maxSearchResults: 500,
         maxUploadSize: process.env.MAX_UPLOAD_SIZE || 10485760,
         autoCreateAndAllocateEnabled: configuration.autoCreateAndAllocateEnabled,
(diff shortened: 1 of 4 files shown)


server/tenantConfig.js
```diff
@@ -38,7 +38,7 @@ async function layoutConfig() {
         maxSearchResults: 500,
         maxUploadSize: process.env.MAX_UPLOAD_SIZE || 10485760,
         autoCreateAndAllocateEnabled: configuration.autoCreateAndAllocateEnabled,
-        defaultTimeoutSeconds: isNaN(defaultTimeoutSeconds) ? 300 : defaultTimeoutSeconds,
+        defaultTimeoutSeconds: isNaN(defaultTimeoutSeconds) ? 1200 : defaultTimeoutSeconds,
```
hubert-raich: Is this ok to go in?

lewismuirhead: Yeah - not really related to this PR but it just helps provide a more pleasant dev experience when there are not any session cookies

PR #419, opened 2020-01-30
HOCS-1462 - Timeline is not ordered to milliseconds precision

hubert-raich:


package-lock.json
```diff
@@ -17228,7 +17228,7 @@
         },
         "os-locale": {
           "version": "1.4.0",
-          "resolved": "https://registry.npmjs.org/os-locale/-/os-locale-1.4.0.tgz",
+          "resolved": "http://registry.npmjs.org/os-locale/-/os-locale-1.4.0.tgz",
```
MarkWrightBjss: Is this correct? Not sure what it's used for.

hubert-raich: just dependency stuff cased by npm install. I've seen Lewis pushing this file out before so it must be fine:)

MarkWrightBjss: I like your logic.

PR #421, opened 2020-02-04
HOCS-1298 : add brought forward date, handle no date value

MarkWrightBjss:
Add brought forward date, handle no date value (was displaying "NaN/NaN/NaN").

### DIFF
--- a/src/shared/common/components/workstack.jsx
+++ b/src/shared/common/components/workstack.jsx
@@ -39,6 +39,8 @@ const ColumnRenderer = {
 
 const dataAdapters = {
     localDate: (value) => {
+        if (!value || value == '')
+            return '';
         var date = new Date(value);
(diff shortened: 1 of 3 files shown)


src/shared/common/components/workstack.jsx
```diff
@@ -39,6 +39,8 @@ const ColumnRenderer = {
 
 const dataAdapters = {
     localDate: (value) => {
+        if (!value || value == '')
+            return '';
         var date = new Date(value);
```
lewismuirhead: looks fine but you could also account for invalid date strings with something like 
```suggestion
        var date = new Date(value);
        if (isNaN(date)) ....
```

MarkWrightBjss: That seems to have worked. Who knew!?!

PR #422, opened 2020-02-05
HOCS-1473 added address details to stage summary

lewismuirhead:


src/shared/common/components/stage-summary.jsx
```diff
@@ -100,9 +101,18 @@ class StageSummary extends Component {
...
                                     <th className='govuk-table__header padding-left--small govuk-!-width-one-third'>Primary correspondent</th>
-                                    <td className='govuk-table__cell'>{summary.primaryCorrespondent}</td>
+                                    <td className='govuk-table__cell'>
+                                        <span>{primaryCorrespondent.fullname}</span>
+                                        {primaryCorrespondent.address && <>
+                                            <br /> <span>{primaryCorrespondent.address.address1}</span>
+                                            <br /> <span>{primaryCorrespondent.address.address2}</span>
+                                            <br /> <span>{primaryCorrespondent.address.address3}</span>
+                                            <br /> <span>{primaryCorrespondent.address.country}</span>
+                                            <br /> <span>{primaryCorrespondent.address.postcode}</span>
```
MarkWrightBjss: Change to match sequence used in UI. Are the labels not required? Also, how would it appear if e.g. "address3" was null?

lewismuirhead: Good point. I was assuming mandatory address lines but they're not.  Not sure it's necessary to label each address line separately for display. The whole address could be in separate row, but again not sure it needs to be

PR #451, opened 2020-05-07
HOCS-1695: Modify adapters/workstacks.js to return columns based on c…

parth-ho:
…onfig & fix corrosponding tests.


server/middleware/searchHandler.js
```diff
@@ -37,11 +37,14 @@ async function handleSearch(req, res, next) {
         const workstackData = await Promise.all(response.data.stages
             .sort(sortObjectByProp(workstackCase => workstackCase.caseReference))
             .map(bindDisplayElements(fromStaticList)));
-        const { workstackColumns } = await req.listService.fetch('S_SYSTEM_CONFIGURATION');
+        const { workstackTypeColumns } = await req.listService.fetch('S_SYSTEM_CONFIGURATION');
+        // using DEFAULT columns for search
+        const workstackColumnsForSearch = workstackTypeColumns[0].workstackColumns;
```
hubert-raich, line 42: Good enough for now... would be nice we we had a dedicated 'SEARCH_WORKSTACK' or similar here. Even if the columns would be the same as 'DEFAULT' for now. 
Saying this, it's very easy to change and improve in the future. I like how it turned out here.

src/shared/common/components/__tests__/__snapshots__/workstack.spec.jsx.snap
```diff
@@ -357,49 +357,49 @@ exports[`Workstack component should render with filtering 1`] = `
...
-              "isFilterable": true,
+              "filterable": true,
               "renderer": "caseLink",
             },
             Object {
               "dataAdapter": null,
               "dataValueKey": "fullName",
               "displayName": "Full Name",
-              "isFilterable": true,
+              "filterable": true,
```
hubert-raich, line 367: I suspect these issues were caused by return the entity 'WorkstackColumn' within 'WorkstackTypeDto' rather than 'WorkstackColumnDto'. I think once we apply the with within info service we can revert these changes here.

src/shared/common/components/workstack.jsx
```diff
@@ -92,7 +92,7 @@ class WorkstackAllocate extends Component {
         const { selectedCases } = this.state;
         if (filter !== '') {
             const filterableColumns = columns.filter(column => {
-                return column.isFilterable === true;
+                return column.filterable === true;
```
hubert-raich, line 95: Same as above, i suspect once we fix the returning objects in info service. This part can be reverted back to 'isFilterable'.

PR #468, opened 2020-05-26
HOCS-1751

Paulcr-ho:
Updating CI scripts with new tokens for the HOCS service separation.
We no longer have everything under CS. They are separated into namespaces.


.drone.yml
```diff
@@ -121,20 +124,59 @@ pipeline:
...
         - poise_whitelist
       commands:
         - cd kube-hocs-frontend
         - ./deploy.sh
       when:
         event: deployment
         environment: prod
-  
-  
+        kube_namespace: cs-prod
```
zuzak, line 163: Don't forget to update the release documentation!

Paulcr-ho: The release process and commands do not change.

PR #485, opened 2020-06-17
HOCS-1804 - MPAM cases are no longer getting overall deadline date

hubert-raich:


zuzak:
Can you amend the commit message to explain why you're making the change?

hubert-raich:
That has messed up the commit history but I have put in more details and also renamed one of the variables to make is slightly less confusing.

Thanks

Hubert

zuzak:
I've rebased those commits for you. Could you double check that I didn't lose any changes in doing so?

src/shared/common/components/stage-summary.jsx
```diff
@@ -119,7 +119,7 @@ class StageSummary extends Component {
                                 {summary.additionalFields && summary.additionalFields.map(({ label, value }) => this.renderRow({ label, value }))}
                             </tbody>
                         </table>
-                        {summary.deadlinesEnabled && <table className='govuk-table margin-left--small'>
+                        {summary.deadlinesEnabled && summary.deadlines && <table className='govuk-table margin-left--small'>
```
MarkWrightBjss, line 122: I think "summary.deadlines" already incorporates "summary.deadlinesEnabled" making "deadlinesEnabled" redundant here.

MarkWrightBjss:
I prefer the new "stageDeadlineEnabled" name - much clearer.

hubert-raich:
HOCS-1804 MPAM cases are no longer getting overall deadline date

Moving summaryDeadlineEnabled so that id does not impact the overall
dedlineEnabled check which is also used to display the overall Case Deadline.
'summaryDeadlineEnabled' should only hide the stage deadlines when false.

I've also added a check on stage-summary.jsx to make sure the heading 'Stage deadlines' is only visible 
when we have at least 1 stage deadline to display

src/shared/common/components/stage-summary.jsx
hubert-raich: Done:) good spot!

PR #487, opened 2020-06-18
HOCS-1808 - Wcs search result page is always blank

hubert-raich:
Updating the fetch logic for 'DEFAULT' workstack type, as it is not guaranteed to always be the first item in the list.

### DIFF
--- a/server/middleware/__tests__/searchHandler.spec.js
+++ b/server/middleware/__tests__/searchHandler.spec.js
@@ -51,8 +51,8 @@ describe('handleSearch', () => {
                     if (listId === 'S_SYSTEM_CONFIGURATION') {
                         return Promise.resolve({
                             workstackTypeColumns: [
-                                { workstackColumns: {} },
-                                { workstackColumns: {} }
+                                { workstackType: 'DEFAULT', workstackColumns: {} },
+                                { workstackType: 'TypeB', workstackColumns: {} }
                             ]
                         }
                         );
--- a/server/middleware/searchHandler.js
+++ b/server/middleware/searchHandler.js
@@ -43,12 +43,14 @@ async function handleSearch(req, res, next) {
             .map(bindDisplayElements(fromStaticList)));
         const { workstackTypeColumns } = await req.listService.fetch('S_SYSTEM_CONFIGURATION');
         // using DEFAULT columns for search
-        const workstackColumnsForSearch = workstackTypeColumns[0].workstackColumns;
+        const workstackTypeForSearch = workstackTypeColumns.find(
+            item => item.workstackType === 'DEFAULT'
+        );
 
         res.locals.workstack = {
             label: 'Search Results',
             items: workstackData,
-            columns: workstackColumnsForSearch
+            columns: workstackTypeForSearch.workstackColumns
         };
 
         next();


server/middleware/__tests__/searchHandler.spec.js
```diff
@@ -51,8 +51,8 @@ describe('handleSearch', () => {
                     if (listId === 'S_SYSTEM_CONFIGURATION') {
                         return Promise.resolve({
                             workstackTypeColumns: [
-                                { workstackColumns: {} },
-                                { workstackColumns: {} }
+                                { workstackType: 'DEFAULT', workstackColumns: {} },
+                                { workstackType: 'TypeB', workstackColumns: {} }
```
zuzak, line 55: Is the naming `TypeB` arbitrary? If so, a more self-explanatory name might be better: or at least one that matches the CAPITALISED_SNAKE_CASE of everything else. What differentiates Type B from the default? That would be a good starting point for coming up with a more meaningful, but still project independent, name.

server/middleware/searchHandler.js
```diff
@@ -43,12 +43,14 @@ async function handleSearch(req, res, next) {
             .map(bindDisplayElements(fromStaticList)));
         const { workstackTypeColumns } = await req.listService.fetch('S_SYSTEM_CONFIGURATION');
         // using DEFAULT columns for search
-        const workstackColumnsForSearch = workstackTypeColumns[0].workstackColumns;
+        const workstackTypeForSearch = workstackTypeColumns.find(
+            item => item.workstackType === 'DEFAULT'
```
zuzak, line 47: ✅

zuzak:
It would be good if you were to amend the Git commit message to explain [what and why](https://chris.beams.io/posts/git-commit/#why-not-how) you changed what you changed. At some point a developer is going to `git blame` this commit to figure out what `TypeB` means.

(As a more minor formatting point, try not to make your commit messages be on one really long line, and give the computer instructions [imperative like "Update the logic", not indicative like "Updating the logic" or "Updated the logic"].)

server/middleware/__tests__/searchHandler.spec.js
hubert-raich: As you can see, there is no difference between the two test workstack definitions. I just needed one to be marked as 'DEFAULT'. I'm happy to change the other definition to 'Other' it doesn't mean anything.  Are you happy with it or have another suggestion?

server/middleware/searchHandler.js
zuzak: Is there a situation in which we will not find a `DEFAULT` workstack? Previously we'd just pick the top workstack, and so this would only break if there were no workstacks at all. Do we need a fallback in case `DEFAULT` isn't found? Or is `{columns: null}` handled elsewhere?

server/middleware/__tests__/searchHandler.spec.js
zuzak: Ah -- I hadn't noticed this was a test! In which case I guess it's fine, unless you feel like writing a test to check it does actually search `DEFAULT` over `TypeB`. I guess you could go for "Some Other Workstack Type" or something.

server/middleware/searchHandler.js
hubert-raich: We have two systems, one of them (WCS) didn't have the 'DEFAULT' workstack defined which I have fixed. So the simple answer is no, all systems have and should have 'DEFAULT' workstack defined otherwise this bug will come back :)

zuzak: I wonder if it's worth throwing an error or asserting that assumption some way, so that if we add a third system and don't add a `DEFAULT` it's very obvious. Otherwise, :shipit:

PR #491, opened 2020-06-26
HOCS-1819 - Hide People's tab on WCS system

hubert-raich:
- suppressed people's tab for wcs cases (logic based on caseId)

### DIFF
--- a/src/shared/common/components/side-bar.jsx
+++ b/src/shared/common/components/side-bar.jsx
@@ -40,6 +40,13 @@ class SideBar extends Component {
     }
 
     render() {
+        let peopleTabEnabled = true;
+        const { page } = this.props;
+        // last chars in caseId always correspond to case_type short code
+        // this check disabled People tab for WCS cases
+        if (page && page.params && page.params.caseId) {
+            peopleTabEnabled = !page.params.caseId.endsWith('c1');
+        }
         return (
             <Fragment >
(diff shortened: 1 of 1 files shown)


src/shared/common/components/side-bar.jsx
```diff
@@ -63,7 +70,7 @@ class SideBar extends Component {
 SideBar.propTypes = {
     activeTab: PropTypes.string,
     page: PropTypes.object.isRequired,
-    track: PropTypes.func.isRequired
+    track: PropTypes.func.isRequired,
```
zuzak, line 73: This change should be in a different commit.

src/shared/common/components/side-bar.jsx
```diff
@@ -40,14 +40,21 @@ class SideBar extends Component {
     }
 
     render() {
+        let peopleTabEnabled = true;
+        const { page } = this.props;
+        // last chars in caseId always correspond to case_type short code
+        // this check disabled People tab for WCS cases
+        if (page && page.params && page.params.caseId) {
+            peopleTabEnabled = !page.params.caseId.endsWith('c1');
```
zuzak, line 48: Are there any scenarios that we could have a case ID that ends with `c1` but isn't WCS, now or in the future? Is there a narrower thing we can match?

PR #494, opened 2020-07-03
Set localstack image version so localstack services remain compatible.

parth-ho:
The localstack image version has been set to `0.9.5`; this localstack services from this version are known to work with all the HOCS/WCS services, whereas a later localstack version broke the search functionality as the version of elasticsearch (from localstack) was not compatible with `hocs-search` service.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -50,7 +50,7 @@ services:
     - ${PWD}/keycloak:/tmp
 
   localstack:
-    image: localstack/localstack:latest
+    image: localstack/localstack:0.9.5
     ports:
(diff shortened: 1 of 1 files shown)


zuzak:
Why did you pick `0.9.5` as the latest compatible version?

parth-ho:
`0.9.5` that was the version on Hubert's machine. Given he's been on the project for a fair while with that version, I reckon any localstack service issues would've come up, so it made sense to go with that.

zuzak:
Okay, fair enough. At some point we should check what the ACP versions are, so that things break before they get deployed.

PR #499, opened 2020-07-27
HOCS-1838: Apply responsive css to specific components.

parth-ho:
Applied responsive css to three entity list components: 1.Correspondents, 2.DraftDocuments, 3.Topics.

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -381,7 +381,7 @@ a.card__body {
             padding: 0.25em 0.5em 0.25em 0.5em;
           }
       }
-      &.indicator {          
+      &.indicator {
           span {
             background-color: black;
             border-radius: 5px;
@@ -400,7 +400,7 @@ a.card__body {
 
   td,th {
     padding-left: 0.2em;
-    
+
     &.govuk-link {
         cursor: pointer;
     }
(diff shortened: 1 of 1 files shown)


src/styles/app.scss
```diff
@@ -761,19 +761,22 @@ $border-colour: govuk-colour("grey-2");
...
-      table-layout: fixed;
-    }
-  }
-}
+#Correspondents, #DraftDocuments, #Topics {
+  table {
+    table-layout: fixed;
+    overflow-wrap: break-word;
+
+    > tbody > tr > td {
```
zuzak, line 769: Would just `td` be sufficient? (And more efficient)

src/styles/app.scss
```diff
@@ -381,7 +381,7 @@ a.card__body {
             padding: 0.25em 0.5em 0.25em 0.5em;
           }
       }
-      &.indicator {          
+      &.indicator {
```
zuzak, line 384: Ideally these whitespace fixes would be in a separate commit

src/styles/app.scss
Paulcr-ho: Verified #Correspondents, #DraftDocuments, #Topics against the form data (loaded in hocs-data project).
It all looks good to me. The govuk-form-group and govuk-table__cell are placed here by the new formatting. 
There is no danger of affecting other components or formatting as they were both inserted in the css for this ticket.

PR #509, opened 2020-08-10
HOCS-1756

hubert-raich:
- added wiring for MPAM campaigns
- type-ahead widget now supports flat key-value input (default format
for entity list values)
- defaultOptions property added to type-ahead widget, when true
all options are visible on blank entry

### DIFF
--- a/src/shared/common/forms/type-ahead.jsx
+++ b/src/shared/common/forms/type-ahead.jsx
@@ -168,7 +168,7 @@ TypeAhead.defaultProps = {
     clearable: true,
     disabled: false,
     choices: [],
-    defaultOptions: false,
+    defaultOptions: false
 };
 
 export default TypeAhead;
--- a/src/shared/common/forms/type-ahead.jsx
+++ b/src/shared/common/forms/type-ahead.jsx
@@ -6,7 +6,19 @@ class TypeAhead extends Component {
 
     constructor(props) {
         super(props);
-        const choices = Array.from(props.choices);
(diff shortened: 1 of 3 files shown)


src/shared/common/forms/type-ahead.jsx
```diff
@@ -146,13 +160,15 @@ TypeAhead.propTypes = {
     label: PropTypes.string,
     name: PropTypes.string.isRequired,
     updateState: PropTypes.func.isRequired,
-    value: PropTypes.string
+    value: PropTypes.string,
+    defaultOptions: PropTypes.bool,
```
zuzak, line 164: hanging comma

src/shared/common/forms/type-ahead.jsx
```diff
@@ -83,6 +96,7 @@ class TypeAhead extends Component {
                     error={error}
                     onChange={this.handleChange.bind(this)}
                     loadOptions={this.getOptions.bind(this)}
+                    defaultOptions={defaultOptions}
```
zuzak, line 99: is this okay to be `null`? (probably)

hubert-raich: Yes, in fact only one widget has it defined. Both undefined and false would mean the same thing in this instance.

src/shared/common/forms/type-ahead.jsx
hubert-raich: fix pending

PR #517, opened 2020-09-25
HOCS-1992 Add case ref to w'stack checkbox labels

zuzak:
Previously we had an empty label element next to the big checkboxes in
workstack view. These labels were required to make the checkboxes render
properly. However, because the labels were empty, they didn't help
screen-readers to understand what the checkboxes were associated with.

This commit adds the case's reference (e.g. HOCS/12345/20) to the
contents of the label, and wraps it in a CSS class that hides it
visually (as it's only useful to screen-readers).

This now means that VoiceOver announces the case reference when you tab
into the associated checkbox, instead of having to tab to the next
column and back again.

This commit fixes an issue that is listed in the DECS accessibility
statement. When this commit gets deployed to production, the
accessibility statement should be updated to reflect that this bug is
fixed.

### DIFF
--- a/src/shared/common/components/workstack.jsx
+++ b/src/shared/common/components/workstack.jsx
@@ -215,7 +215,9 @@ class WorkstackAllocate extends Component {
                                 onChange={handleChange.bind(this)}
                                 className={'govuk-checkboxes__input'}
                             />
-                            <label className='govuk-label govuk-checkboxes__label' htmlFor={`selected_cases_${item.caseUUID}`}></label>
+                            <label className='govuk-label govuk-checkboxes__label' htmlFor={`selected_cases_${item.caseUUID}`}>
+                                <span className='govuk-visually-hidden'>{item.caseReference}</span>
+                            </label>
                         </div>
                     </div>
                 </td>}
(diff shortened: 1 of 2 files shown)


src/shared/common/components/workstack.jsx
```diff
@@ -215,7 +215,9 @@ class WorkstackAllocate extends Component {
                                 onChange={handleChange.bind(this)}
                                 className={'govuk-checkboxes__input'}
                             />
-                            <label className='govuk-label govuk-checkboxes__label' htmlFor={`selected_cases_${item.caseUUID}`}></label>
+                            <label className='govuk-label govuk-checkboxes__label' htmlFor={`selected_cases_${item.caseUUID}`}>
+                                <span className='govuk-visually-hidden'>{item.caseReference}</span>
```
parth-ho, line 219: Wouldn't a `aria-label` property achieve the same thing, without having to nest a span element?

zuzak: I'm not an expert here, but it looks like `aria-label` is there for when there's no associated label element, whereas here we have a label element that's blank?

I'm not sure _why_ we need a label element here, but without it the checkbox loses its big GOV.UK styling.

parth-ho (approved):

PR #523, opened 2020-10-07
HOCS-1960: add case note for send to workflow manager

LiamMacP:
Add the title and pass the note for notes that come through with case type of SEND_TO_WORKFLOW_MANAGER. Add flag for it to be classed as a case note on the timeline item.

### DIFF
--- a/src/shared/common/components/case-notes.jsx
+++ b/src/shared/common/components/case-notes.jsx
@@ -41,7 +41,7 @@ AuditEvent.propTypes = {
 };
 
 const TimelineItem = (refreshNotes) => ({ type, body, title, timelineItemUUID }) => {
-    const isCaseNote = ['MANUAL', 'ALLOCATE', 'CLOSE', 'REJECT', 'PHONECALL', 'REQUEST_CONTRIBUTION', 'FOLLOW_UP', 'FOLLOW_UP_NOT_COMPLETED', 'WITHDRAW'].includes(type);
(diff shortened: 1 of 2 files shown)


src/shared/common/components/case-notes.jsx
```diff
@@ -41,7 +41,7 @@ AuditEvent.propTypes = {
 };
 
 const TimelineItem = (refreshNotes) => ({ type, body, title, timelineItemUUID }) => {
-    const isCaseNote = ['MANUAL', 'ALLOCATE', 'CLOSE', 'REJECT', 'PHONECALL', 'REQUEST_CONTRIBUTION', 'FOLLOW_UP', 'FOLLOW_UP_NOT_COMPLETED', 'WITHDRAW'].includes(type);
+    const isCaseNote = ['MANUAL', 'ALLOCATE', 'CLOSE', 'REJECT', 'PHONECALL', 'REQUEST_CONTRIBUTION', 'SEND_TO_WORKFLOW_MANAGER', 'FOLLOW_UP', 'FOLLOW_UP_NOT_COMPLETED', 'WITHDRAW'].includes(type);
```
hubert-raich, line 44: At some point here maybe it will be better to figure out what all non-case-note types we have and invert this check.
This would reduce the need for keeping and maintaining this list. Not a must do for this change as you're just following existing pattern :)

hubert-raich (approved):
All good

src/shared/common/components/case-notes.jsx
LiamMacP: I'd also try and find a wait of making these into some sort of enum, I really do not like the use of hardcoded strings like this when we might need to use them elsewhere in the future...

PR #524, opened 2020-10-11
HOCS-2040: change to use 24 hour clock

LiamMacP:
Previously when looking at case notes in the timeline we showed the 12 hour clock. This could get unnecessarily confusing as a case note that is added at 7:31pm would just show 7:31 without the meridiem at the end, meaning the user could be confused as to whether this is AM/PM.

This update toggles the International Date Format to use 24 hour time so the previous example would show 19:31 instead of 7:31.

### DIFF
--- a/server/lists/adapters/case-notes.js
+++ b/server/lists/adapters/case-notes.js
@@ -126,7 +126,8 @@ const formatDate = (rawDate) => {
         day: 'numeric',
         hour: 'numeric',
         minute: 'numeric',
-        timeZone: 'Europe/London'
+        timeZone: 'Europe/London',
+        hour12: false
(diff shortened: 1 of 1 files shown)


zuzak (approved):
Personally I'd've put the new line above the `timeZone` line to reduce the amount of code changes in the diff, but that's me being particularly finickety. Don't worry about it here.

Incidentally, if you explicitly set `hour12: true` it'll add the am/pm. The [style guide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style) for GOV.UK (which isn't binding on us) would prefer "5:30pm" instead of "17:30" or "5:30 pm", but really it depends on which our users tend to think in.

PR #525, opened 2020-10-12
HOCS-2038: expandable checkbox no longer saves undefined

LiamMacP:
When unchecking a already selected expandable checkbox, previously this would save undefined as the value. Due to this the front-end summary page and the audit extracts show this value as undefined. This is not anticipated behaviour. This change fixes this by saving empty as the value, thus, fixing the summary page and the audit extraction.


abarker66 (approved):
Code explained and test updated.

PR #527, opened 2020-10-16
added two new case note type adaptors

abarker66:
Additions for mandatory case notes when converting a case from ministerial to official or vice versa.


LiamMacP (approved):
Changes checked and approved. 

Just a note, hocs-frontend uses 'Feature' as the branching naming strategy and not 'feature'. As a result of how the repo is set up for case insensitive branching, people cannot pull 'feature' branches.

PR #529, opened 2020-10-23
HOCS-2096: add further logging

LiamMacP:
Move up the logging of the case reference to just after retrieving from the request for production issue. Then add further logging to capture encoded string.

### DIFF
--- a/server/routes/dashboard.js
+++ b/server/routes/dashboard.js
@@ -25,6 +25,9 @@ router.post(['/search/reference', '/api/search/reference'],
         const logger = getLogger(req.requestId);
         try {
             const caseRef = req.form.data['case-reference'].toUpperCase();
+
+            logger.info('SEARCH_REFERENCE', { reference: caseRef });
+
             const reference = encodeURIComponent(caseRef);
 
             const response = await caseworkService.get(`/case/${reference}/stage`, {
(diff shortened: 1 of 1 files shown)


server/routes/dashboard.js
```diff
@@ -25,14 +25,18 @@ router.post(['/search/reference', '/api/search/reference'],
         const logger = getLogger(req.requestId);
         try {
             const caseRef = req.form.data['case-reference'].toUpperCase();
+
+            logger.info('SEARCH_REFERENCE', { reference: caseRef });
+
             const reference = encodeURIComponent(caseRef);
 
+            logger.info('SEARCH_REFERENCE_ENCODED', { encoded_reference: reference });
```
zuzak, line 33: Could this be consolidated into a single line of logging? (I don't know much about our logging ecosystem)
```suggestion
            logger.info('SEARCH_REFERENCE_ENCODED', { encoded_reference: reference, reference: caseRef });
```

LiamMacP: I only added it twice incase for some reason the encoding goes bang. Shouldn't but always a chance?

zuzak: But [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) is a relatively uncontroversial and predictable bit of code, so is this extra bit of logging useful?

LiamMacP: Yeah I agree 👍🏼

LiamMacP: Changed to remove this unnecessary logging.

server/routes/dashboard.js
```diff
@@ -25,14 +25,16 @@ router.post(['/search/reference', '/api/search/reference'],
         const logger = getLogger(req.requestId);
         try {
             const caseRef = req.form.data['case-reference'].toUpperCase();
+
```
zuzak, line 28: if it were me I'd remove this new line to keep the "caseRef declaration stuff" in one block. But that's all personal preference.

zuzak (approved):

PR #531, opened 2020-10-26
HOCS-2004: Handle multiple data value keys for workstack

LiamMacP:
As a result of requested change by WCS, the capability to have multiple data_value_keys has been added. This allows the user to specify a data_value_key and if it is present it will show. If the key is not present it will then fallback to the next and so forth. 

Tests have been updated to resemble this new functionality.

### DIFF
--- a/src/shared/common/components/__tests__/__snapshots__/workstack.spec.jsx.snap
+++ b/src/shared/common/components/__tests__/__snapshots__/workstack.spec.jsx.snap
@@ -408,14 +408,14 @@ exports[`Workstack component should render with filtering 1`] = `
             },
             Object {
               "dataAdapter": "localDate",
-              "dataValueKey": "created",
+              "dataValueKey": "TEST,created",
               "displayName": "Case Date",
               "isFilterable": true,
               "renderer": null,
             },
             Object {
               "dataAdapter": "indicator",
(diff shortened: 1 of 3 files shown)


src/shared/common/components/__tests__/__snapshots__/workstack.spec.jsx.snap
```diff
@@ -156,7 +156,7 @@ Array [
                       <td
                         class="govuk-table__cell"
                       >
-                        is Active
+                        is Active,TEST
```
zuzak, line 159: Are these meant to be here?

LiamMacP: Yes when we specify a renderer of indicator and the value is YES.

LiamMacP: Thinking about it, it shouldn't, it should only be the active thats shown as its matched the first. Will rectify tomorrow.

LiamMacP: This has been fixed, thanks for pointing it out 👍🏼

Paulcr-ho (approved):
Suggested change made.

PR #533, opened 2020-10-28
HOCS-1990: remove unnecessary ES file

LiamMacP:
As a result of the elastic search index being moved to our private repo, we no longer have a need to instantiate with this empty file. General cleanup.


Paulcr-ho (approved):
We will have to include the removed step in our documentation for the data projects (where the index mapping files are)

PR #543, opened 2020-11-19
HOCS-2163: Enforce stricter case reference regex

LiamMacP:
At present when searching by a case reference we check to see if the field contains '\b[a-zA-Z]{2,4}\/[0-9]{7}\/[0-9]{2}\b'. This regex checks the entire field text and passes if it contains the basic case reference pattern. This is not acceptable though as you can have anything else either next to it/ before it that would also pass, 'test MIN/0000000/00' would still be valid. This regex has been updated to enforce that the field matches exactly with the case reference pattern.

Alongside this we are currently not in line with what the error message style guide states we should return as incorrect input. Before we had improper english and used the word invalid. This commit changes this to inform the user they have to enter a case reference in the correct format.

At present we don't have any tests for the isValidCaseReference validator. This adds tests that look for invalid formats, alongside checks that a valid one can be case insensitive.

### DIFF
--- a/server/middleware/__tests__/validation.spec.js
+++ b/server/middleware/__tests__/validation.spec.js
@@ -148,6 +148,29 @@ describe('Validators', () => {
         });
     });
 
+    describe('Case reference validation', () => {
+        it('should accept regardless of case', () => {
+            expect(validators.isValidCaseReference({ value: 'MiN/0000000/22' })).toEqual(null);
+        });
+        it('should accept spaces at the end', () => {
+            expect(validators.isValidCaseReference({ value: 'MIN/0000000/22 ' })).toEqual(null); // one space at the end
+            expect(validators.isValidCaseReference({ value: ' MIN/0000000/22' })).toEqual(null); // one space at the beginning
+            expect(validators.isValidCaseReference({ value: 'MIN/0000000/22     ' })).toEqual(null); // multiple spaces at the end
+            expect(validators.isValidCaseReference({ value: '     MIN/0000000/22' })).toEqual(null); // multiple spaces at the beginning
+        });
+        it('should reject if does not fit pattern', () => {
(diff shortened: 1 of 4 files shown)


zuzak (commented):
You've changed the RegExp from accepting ` MIN/1234567/89 ` to just `MIN/1234567/89`, and are forcing the user to notice they've pasted a difficult-to-notice space after their case reference.

Ideally we would ahere to [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) and `.trim()` instead of forcing the user to remove the whitespace themselves. We're meant to [do the hard work to make it simple](https://www.gov.uk/guidance/government-design-principles#do-the-hard-work-to-make-it-simple): fixing the problem this way might save one-off developer time, but there are many more users than developers and they will be performing this action multiple times a day.

You might want to add additional test cases for having the wrong number of digits, or the wrong number of letter prefixes.

Paulcr-ho (dismissed):

LiamMacP:
> You've changed the RegExp from accepting `MIN/1234567/89` to just `MIN/1234567/89`, and are forcing the user to notice they've pasted a difficult-to-notice space after their case reference.
> 
> Ideally we would ahere to [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) and `.trim()` instead of forcing the user to remove the whitespace themselves. We're meant to [do the hard work to make it simple](https://www.gov.uk/guidance/government-design-principles#do-the-hard-work-to-make-it-simple): fixing the problem this way might save one-off developer time, but there are many more users than developers and they will be performing this action multiple times a day.
> 
> You might want to add additional test cases for having the wrong number of digits, or the wrong number of letter prefixes.

This has now been rectified 👍🏼

Paulcr-ho (approved):

PR #545, opened 2020-11-26
HOCS-2222: change the case type topics to dynamic

LiamMacP:
At present the case types are loaded as a static list, this is an issue as we allow for topics to be added to teams through MUI. This change forced the list to be dynamic, which means that every time they are needed they are loaded in. This will allow new cases that are assigned to teams to show in the core application.

### DIFF
--- a/server/lists/index.js
+++ b/server/lists/index.js
@@ -296,6 +296,7 @@ module.exports = {
         TOPICS_CASETYPE: {
             client: 'INFO',
             endpoint: '/case/${caseId}/topiclist',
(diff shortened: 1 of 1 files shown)


zuzak:
Are there performance implications here?

Paulcr-ho (approved):

LiamMacP:
> Are there performance implications here?

Yes there will be, anytime someone from DCU is to add a topic to a case, it will refresh this list. 

At the minute we don't have any performance testing that is run against our code base, this is definitely something we need to start doing.

PR #546, opened 2020-11-26
HOCS-2226: Update govuk-frontend package from 2.9.0 to 3.10.0

parth-ho:
note: The guidance I used to replace the old colours is here: https://frontend.design-system.service.gov.uk/migrating-from-legacy-products/#2-turn-off-compatibility-mode

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
         "cookie-parser": "^1.4.4",
         "csurf": "^1.10.0",
         "express": "^4.16.4",
-        "govuk-frontend": "2.9.0",
+        "govuk-frontend": "3.10.0",
         "jsonwebtoken": "^8.5.1",
         "multer": "^1.4.2",
         "multer-s3": "^2.7.0",
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -6,7 +6,7 @@ $govuk-page-width: 1300px !default;
 $govuk-font-family: 'helvetica neue', Arial, Helvetica, sans-serif !default;
 $govuk-font-family-tabular: 'helvetica neue',Arial, Helvetica, sans-serif !default;
 
-@import "govuk-frontend/all.scss";
+@import "node_modules/govuk-frontend/govuk/all";
 
 @function tint($color, $percentage) {
   @return mix(white, $color, $percentage);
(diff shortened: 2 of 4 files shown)


src/styles/app.scss
```diff
@@ -57,7 +57,7 @@ $govuk-font-family-tabular: 'helvetica neue',Arial, Helvetica, sans-serif !defau
   padding: 0 govuk-spacing(1);
   overflow-x: auto;
   color: govuk-colour("pink");
-  background-color: govuk-colour("grey-4");
+  background-color: govuk-colour("light-grey");
```
zuzak, line 60: This fails the WCAG 2.1 contrast guidelines (https://webaim.org/resources/contrastchecker/).
(The old colour did too, mind)

zuzak (commented):
The [GOV.UK Design System release-notes](https://github.com/alphagov/govuk-frontend/releases/tag/v3.0.0) for v3.0.0 mention a few things that this commit doesn't fix. For example, we use the `data-module` attribute for our accordions: if we're using the GDS code for that, we'll need to update it here too.

I remember vaguely looking into this ticket a while ago and putting it on the backburner because there were lots of small changes that I didn't have time for.

parth-ho:
@zuzak I'm not sure how much we should rely on the release notes for this upgrade, that's just what they've broadcasted, and we don't know how comprehensive these notes are. We also have lots of react and custom JS code interacting with the components too. Might be better and more practical to run some comprehensive tests. If all tests still work then I would be confident that everything is ok.

zuzak:
The GDS release notes tend to be comprehensive for each release, but are of course are targeted at more vanilla users of the Design System than us. Looks like that data-module stuff is in our accordions, but as a result of blind copy/pasting -- it doesn't seem to actually do anything.

We should also add a ticket to the backlog to go through the changes from 2.9 to present to see if there are new features we should in turn be implementing.

package.json
```diff
@@ -45,7 +45,7 @@
         "cookie-parser": "^1.4.4",
         "csurf": "^1.10.0",
         "express": "^4.16.4",
-        "govuk-frontend": "2.9.0",
+        "govuk-frontend": "3.10.0",
```
zuzak, line 48: Wonder why this doesn't use caret notation.

zuzak (approved):

package.json
LiamMacP: They do seem to use patch versions, might be wise to change this next time we go into this file.

PR #547, opened 2020-12-04
HOCS-2019 : add new case note type for close case by telephone.

abarker66:
Added a new case note type for the close case by telephone events.


LiamMacP (approved):
Looks good!

LiamMacP (approved):

PR #548, opened 2020-12-07
HOCS-2136: Transfer to OGD at MPAM Creation stage

parth-ho:
This PR contains the functionality to input MPAM missalocations data the data input stage.

### DIFF
--- a/server/middleware/validation.js
+++ b/server/middleware/validation.js
@@ -106,6 +106,26 @@ const validators = {
     }
 };
 
+function validateConditionalRadioContentIfExists(data, name, choices, validator, result) {
+    const conditionalRadioButtonTextFieldId = `${data[name]}Text`;
+
+    if (conditionalRadioButtonTextFieldId in data) {
+        const radioChoice = choices.find(choice => {
+            return choice.value === data[name];
+        });
+        let label = radioChoice.conditionalContent.label;
+        const value = data[conditionalRadioButtonTextFieldId];
+
+        const validationError = validators[validator].call(
+            this,
+            { label, value }
+        );
+        if (validationError) {
+            result[conditionalRadioButtonTextFieldId] = validationError;
+        }
+    }
+}
+
 function validationMiddleware(req, res, next) {
     if (req.form) {
         try {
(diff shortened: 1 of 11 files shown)


server/middleware/validation.js
```diff
@@ -139,24 +159,39 @@ function validationMiddleware(req, res, next) {
...
     next();
 
-    function isFieldVisible(visibilityConditions, data) {
+    function isFieldVisible(visibilityConditions, hideConditions, data) {
+        let i;
         let isVisible = true;
         if (visibilityConditions) {
             isVisible = false;
-            for (var i = 0; i < visibilityConditions.length; i++) {
+            for (i = 0; i < visibilityConditions.length; i++) {
```
zuzak, line 167: I'd put `let i` here (I know if it were `var` JavaScript hoisting would mean it made no difference, but I think here it'd mean the variable disappears when we exit the `for` loop, which would be safer). See `isComponentVisible()` further down this PR.

src/shared/common/forms/form-repository.jsx
```diff
@@ -49,15 +49,28 @@ function renderFormComponent(Component, options) {
...
             if (data[condition.conditionPropertyName] && data[condition.conditionPropertyName] === condition.conditionPropertyValue) {
                 isVisible = true;
             }
         }
     }
+
+    // hide component based on hideConditions
+    if (hideConditions) {
+        for (let i = 0; i < hideConditions.length; i++) {
+            const condition = hideConditions[i];
```
zuzak, line 68: Does JS mind multiple `const` here?

server/middleware/validation.js
```diff
@@ -139,24 +159,39 @@ function validationMiddleware(req, res, next) {
...
                 const condition = visibilityConditions[i];
                 if (data[condition.conditionPropertyName] && data[condition.conditionPropertyName] === condition.conditionPropertyValue) {
                     isVisible = true;
                 }
             }
         }
+
+        if (hideConditions) {
+            isVisible = true;
+            for (i = 0; i < hideConditions.length; i++) {
```
zuzak, line 177: Consider:
```js
for (let condition of hideConditions) {
```

zuzak (commented):
You've added a merge commit to your pull request. Ideally you'd rebase instead, so we don't have references in `master` to branches we shortly don't care about.

You could rebase it away using `git rebase -i origin/master` and changing the word next to that commit to `drop`.

src/shared/common/forms/form-repository.jsx
parth-ho: Seem this is fine: https://stackoverflow.com/questions/50807989/es6-javascript-const-inside-or-let-outside-loop

PR #549, opened 2020-12-11
HOCS-2259: overdue no longer includes current day

LiamMacP:
At present when calculating how many items are overdue we include the current day in the calculation. This is in part due to us only storing the date '2020-12-31' in the database. We then checked that against the current time but when we parse '2020-12-31' is defaults to hour, minute, second and millisecond  of 0. Thus, we are below the current timestamp. This change forces the due date to be at the end of the day.

Tests have been added to make sure we don't regress this in the future.


server/lists/adapters/__tests__/workstacks.spec.js
```diff
@@ -17,6 +17,8 @@ const mockFromStaticList = jest.fn((list) => {
     }
 });
 
+const todaysDate = new Date().toISOString().slice(0,10); // Returns todays date in format (yyyy-MM-dd)
```
zuzak, line 20: Splitting on `T` might be more reliable.

server/lists/adapters/workstacks.js
```diff
@@ -27,7 +27,7 @@ const defaultCaseSort = (a, b) => {
 };
 const byLabel = (a, b) => a.label.localeCompare(b.label);
 const isUnallocated = user => user === null;
-const isOverdue = (configuration, deadline) => configuration.deadlinesEnabled && deadline && new Date(deadline) < Date.now();
+const isOverdue = (configuration, deadline) => configuration.deadlinesEnabled && deadline && new Date(deadline).setHours(23, 59, 59, 999) < Date.now();
```
zuzak, line 30: This is an interesting approach. Would it be better to check if it's after 00:00:00 tomorrow, rather than 23:59:59 today?

The line's also getting a bit long -- I'd add a couple of line-breaks.

LiamMacP: We'd have to do something like this: 
https://stackoverflow.com/a/563442

But that would require more computation, this solution is simpler. 

Will add more line breaks, to simplify the view.

server/lists/adapters/__tests__/workstacks.spec.js
LiamMacP: That would return an array, and we don't care about the time. So slicing by the first 10 characters on a string that's a ISO standard is simpler. If you are adamant on this change I can do it.

LiamMacP: This has been rectified 👍🏼

server/lists/adapters/workstacks.js
LiamMacP: This has been rectified 👍🏼

server/lists/adapters/__tests__/workstacks.spec.js
```diff
@@ -17,7 +17,7 @@ const mockFromStaticList = jest.fn((list) => {
     }
 });
 
-const todaysDate = new Date().toISOString().slice(0,10); // Returns todays date in format (yyyy-MM-dd)
+const todaysDate = new Date().toISOString().split('T')[0]; // Returns todays date in format (yyyy-MM-dd)
```
zuzak, line 20:
```suggestion
onst todaysDate = new Date().toISOString().split('T')[0]; // Returns today's date in format (yyyy-MM-dd)
```

zuzak: Wonder if this should be in dateHelpers as well?

server/libs/dateHelpers.js
```diff
@@ -0,0 +1,7 @@
+function addDays(dateString, days) {
+    var date = new Date(dateString);
```
zuzak, line 2:
```suggestion
    const date = new Date(dateString);
```

server/libs/__tests__/dateHelpers.spec.js
```diff
@@ -0,0 +1,31 @@
...
+        expect(date).toEqual(new Date('2021-01-02'));
+    });
+
+    it('should roll a year over on adding day', () => {
+        const date = addDays('2020-12-31', 1);
+
+        expect(date).toBeDefined;
+        expect(date).toEqual(new Date('2021-01-01'));
+    });
+});
```
zuzak, line 31: This is less important, but ideally there should be line endings at the end of files so that it plays nicer with `git diff` in future (as to add a line 32 you'd have to edit line 31 to stick a line ending on the end).

server/lists/adapters/__tests__/workstacks.spec.js
LiamMacP: This has been moved and tests added 👍🏼

server/libs/__tests__/dateHelpers.spec.js
LiamMacP: Rectified 👍🏼

server/libs/dateHelpers.js
LiamMacP: Rectified 👍🏼

Paulcr-ho (approved):
All requested changes have been made. Approving.

PR #552, opened 2020-12-15
create 'workflow' schema

philbarton:
Tiny change to save manual step to create schema.

### DIFF
--- a/docker/db_setup/workflow.sql
+++ b/docker/db_setup/workflow.sql
@@ -1,0 +1,2 @@
+\c postgres;
+CREATE SCHEMA workflow;
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -12,6 +12,8 @@ services:
     environment:
       POSTGRES_USER: root
       POSTGRES_PASSWORD: dev
+    volumes:
+      - ./db_setup:/docker-entrypoint-initdb.d
     tmpfs:
     - /var/lib/postgresql
 


LiamMacP:
How does this work with the starting the hocs-workflow project without the schema existing? Does it create it? all but hocs-workflow create the schema within the java project if it doesn't exist already.

philbarton:
I think you said that this is only missing locally. This just adds the schema from the front-end docker compose.
Any sql files in that directory get run at postgres startup. (If the volume has been removed)

zuzak:
Boring Git comments:

Would be good if you could rephrase (`git commit --amend -m "HOCS-2264 Create workflow schema" && git push origin +refactor/HOCS-2264-create-workflow-schema`) the git commit so that the Jira ticket is in the first line of the commit.

(and if you're feeling generous, to use the third line of the commit message to explain why you're creating the schema -- we know _what_ you've done, because we can read the code, but it won't be obvious _why_ you did that in six months' time when someone `git annotate`s to chase down a bug).

Also, you'll find GitHub works slightly better if your digital email address is associated to your GitHub account (you can do that at https://github.com/settings/emails).

LiamMacP:
> I think you said that this is only missing locally. This just adds the schema from the front-end docker compose.
> Any sql files in that directory get run at postgres startup. (If the volume has been removed)

Yeah that should fix that edge case. What about this though, we delete the schema after postgres has started, then try to run hocs-workflow. The schema doesn't exists then, and won't be created, which is an issue because the other projects do create it. I'm happy with the change though as it fixes the problem from one angle.

LiamMacP:
@philbarton one more question, if the schema already exists will this fail/ error out? Might be worth prepending the following to ensure that it doesn't create if it already exists: `IF NOT EXISTS`

philbarton:
> Boring Git comments:
> 
> Would be good if you could rephrase (`git commit --amend -m "HOCS-2264 Create workflow schema" && git push origin +refactor/HOCS-2264-create-workflow-schema`) the git commit so that the Jira ticket is in the first line of the commit.
> 
> (and if you're feeling generous, to use the third line of the commit message to explain why you're creating the schema -- we know _what_ you've done, because we can read the code, but it won't be obvious _why_ you did that in six months' time when someone `git annotate`s to chase down a bug).
> 
> Also, you'll find GitHub works slightly better if your digital email address is associated to your GitHub account (you can do that at https://github.com/settings/emails).

Thanks Douglas, my bad. That's the convention I used on my last few projects. No Idea why I didn't do it the....

philbarton:
> > I think you said that this is only missing locally. This just adds the schema from the front-end docker compose.
> > Any sql files in that directory get run at postgres startup. (If the volume has been removed)
> 
> Yeah that should fix that edge case. What about this though, we delete the schema after postgres has started, then try to run hocs-workflow. The schema doesn't exists then, and won't be created, which is an issue because the other projects do create it. I'm happy with the change though as it fixes the problem from one angle.

This schema is third party, owned by camunda. Thats why is doesn't behave like the other services. It must be created in RDS for real deployments some where. I'll take a look out of interest.

philbarton:
> @philbarton one more question, if the schema already exists will this fail/ error out? Might be worth prepending the following to ensure that it doesn't create if it already exists: `IF NOT EXISTS`

Scripts in /docker-entrypoint-initdb.d are only run if you start the container with a data directory that is empty; any pre-existing database will be left untouched on container startup. 
Its neat feature of dockerised DBs https://hub.docker.com/_/postgres

LiamMacP (approved):
Changes look good, informed decision and well thought through.

PR #565, opened 2021-01-08
HOCS-2051 ukvi search by correspondent results page

agraham19:
This change is part of HOCS-2051 to allow UKVI users to filter by correspondent. This PR populates workstack display elements for correspondents

This should be merged in at the same time as the following PRs:
https://github.com/UKHomeOffice/hocs-frontend/pull/565
https://github.com/UKHomeOffice/hocs-data/pull/489
https://github.com/UKHomeOffice/hocs-casework/pull/367

### DIFF
--- a/server/lists/adapters/workstacks.js
+++ b/server/lists/adapters/workstacks.js
@@ -109,6 +109,11 @@ const returnMyCasesWorkstackColumns = (configuration, workstackData) => {
     return getColumnsForMyCases.workstackColumns;
 };
 
+const getCorrespondentsNameByType = (correspondents, types) =>
+    correspondents.correspondents.filter(correspondent => types.includes(correspondent.type))
+        .map(correspondent => correspondent.fullname)
+        .join(', ');
+
 const bindDisplayElements = fromStaticList => async (stage) => {
     stage.assignedTeamDisplay = await fromStaticList('S_TEAMS', stage.teamUUID);
     stage.caseTypeDisplayFull = await fromStaticList('S_CASETYPES', stage.caseType);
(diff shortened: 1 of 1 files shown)


server/lists/adapters/workstacks.js
```diff
@@ -138,10 +144,19 @@ const bindDisplayElements = fromStaticList => async (stage) => {
     }
     stage.deadlineDisplay = formatDate(stage.deadline);
     if (stage.data && stage.data.DueDate) {
-        stage.stageTypeWithDueDateDisplay = stage.stageTypeDisplay + ' due: ' + formatDate(stage.data.DueDate);
+        stage.stageTypeWithDueDateDisplay = stage.stageTypeDisplay + ' due: ' + formatDate(stage.stage.data.DueDate);
```
LiamMacP, line 147: This has heavily changed and we no longer use DueDate as a result of the new MultipleContributions piece.

LiamMacP (changes requested):
Change required as DueDate no longer used. 

Also don't think we need the 'docker/scripts/keycloak/hsperfdata_jboss/125' as is an error file from keycloak.

server/lists/adapters/workstacks.js
agraham19: I have now rebased

agraham19:
> Change required as DueDate no longer used.
> 
> Also don't think we need the 'docker/scripts/keycloak/hsperfdata_jboss/125' as is an error file from keycloak.

I have now rebased

LiamMacP (approved):

PR #568, opened 2021-01-11
HOCS-2285 Fix console errors when not running docs locally

ARobinsonHOD:
- Added logic to handle unexpected (or missing) response data from API call to hocs-docs preventing console errors.
- Separated `flattenDocuments`, `getFirstDocument` and `hasPendingDocuments` into their own helper class so they could be unit tested.
- Add unit tests for expected and unexpected values for those three functions.

### DIFF
--- a/src/shared/common/components/document-pane.jsx
+++ b/src/shared/common/components/document-pane.jsx
@@ -10,7 +10,12 @@ import {
 } from '../../contexts/actions/index.jsx';
 import status from '../../helpers/api-status.js';
 import { Link } from 'react-router-dom';
-import sortByTimeStamp from '../../helpers/sortByTimeStamp.js';
+import {
+    flattenDocuments,
+    getFirstDocument,
+    hasPendingDocuments
+} from '../../helpers/document-helpers';
+
 
 class DocumentPanel extends Component {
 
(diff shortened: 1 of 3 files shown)


src/shared/common/components/document-pane.jsx
```diff
@@ -12,13 +12,42 @@ import status from '../../helpers/api-status.js';
 import { Link } from 'react-router-dom';
 import sortByTimeStamp from '../../helpers/sortByTimeStamp.js';
 
+class DocumentPanelHelpers {
```
LiamMacP, line 15: No real need for this to be a class. 

I'd be happy with these helper being moves to src/shared/helpers under a filename `DocumentPanelHelpers.js`. The functions can then be exported and imported into the component. Therefore we have 
[https://en.wikipedia.org/wiki/Separation_of_concerns](separation of concerns). Tests then could also be moved to src/shared/helpers/__tests__.

LiamMacP: The functions would no longer need to be static either.

LiamMacP (changes requested):

src/shared/common/components/document-pane.jsx
ARobinsonHOD: Thanks @LiamMacP, I've moved them out into a separate file and done the same with the unit tests.

LiamMacP (approved):
Changes look good, full test coverage and ticket has been completed.

PR #569, opened 2021-01-12
HOCS-2233: Escape filenames before uploading to S3.

parth-ho:


zuzak (approved):
Parallel commit of https://github.com/UKHomeOffice/hocs-management-ui/pull/64.

PR #575, opened 2021-01-18
HOCS-2313: contribution received date limit range

LiamMacP:
As per the ticket add the is valid within date validator to the contribution received date. This therefore enforces that the date is within the last 180 days (unless changes through environment variable). This PR also cleans up the contribution fulfilment and contribution request screen by removing unnecessary validation messages.

### DIFF
--- a/server/services/forms/schemas/contribution-request.js
+++ b/server/services/forms/schemas/contribution-request.js
@@ -19,7 +19,7 @@ module.exports = async options => {
         .withTitle(`${isAdd ? 'Add' : isReadOnly ? 'View' : 'Edit'} Contribution Request`)
         .withField(
             Component('dropdown', 'contributionBusinessArea')
-                .withValidator('required', 'Business area is required')
+                .withValidator('required')
                 .withProp('label', 'Business Area')
                 .withProp('disabled', isReadOnly)
                 .withProp('choices', [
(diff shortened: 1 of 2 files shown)


server/services/forms/schemas/contribution-request.js
```diff
@@ -51,25 +51,25 @@ module.exports = async options => {
...
+                .withValidator('isValidDate')
+                .withValidator('isAfterToday')
                 .withProp('label', 'Contribution due date')
                 .withProp('disabled', isReadOnly)
                 .build()
         )
         .withField(
             Component('text-area', 'contributionRequestNote')
-                .withValidator('required', 'What you are requesting is required')
+                .withValidator('required')
```
zuzak, line 72: I think this'll mean it ends up with "Contribution Request Notes is required", which reads strangely.

LiamMacP: Validation is based on the label of the component. In this case as we set the label to be `What you are requesting` on line 73. This will read as 'What you are requesting is required' which is what it said before.

zuzak (approved):
Looks fine to me. Would be good to rebase onto master (instead of a merge).

Your Git commit messages aren't quite right -- ideally you'd [wrap the text at 72 characters](https://chris.beams.io/posts/git-commit/#wrap-72).

PR #579, opened 2021-01-21
HOCS-2322: Reintroduce DueDate to Workstack

LiamMacP:
This PR reintroduces the DueDate check to the workstack component and adapter.
This fixes MC removing DueDate from showing for Dispatch Follow up Cases.

### DIFF
--- a/src/shared/common/components/workstack.jsx
+++ b/src/shared/common/components/workstack.jsx
@@ -288,7 +288,6 @@ class WorkstackAllocate extends Component {
                 }
                 return <td key={row.uuid + column.dataValueKey} className='govuk-table__cell'>{value}</td>;
             case ColumnRenderer.DUE_DATE_WARNING:
-                console.log(row.data);
                 if (row.data.CaseContributions) {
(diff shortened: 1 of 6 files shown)


src/shared/common/components/workstack.jsx
```diff
@@ -302,6 +301,13 @@ class WorkstackAllocate extends Component {
                         </td>;
                     }
                 }
+                if (row.data.DueDate) {
+                    if (new Date(row.data.DueDate) <= new Date()) {
+                        return <td key={row.uuid + column.dataValueKey} className='govuk-table__cell date-warning'>
+                            <span>{value}</span>
```
zuzak, line 307: Is the span required?

LiamMacP: The css fiile does it based on the class dateWarning with a span. This is more of a revert into MC so don't want to make wider changes than required. Also without looking we might apply default span css to TD's. Happy to create a ticket to look at this further in the future.

Paulcr-ho (approved):

PR #583, opened 2021-02-02
HOCS-2395-Remove-broken-aria-reference

agraham19:
The other input fields do not use aria labels, so for consistency:
* remove broken aria-describeby and rely on form label for screenreader description

### DIFF
--- a/src/shared/common/forms/__tests__/__snapshots__/radio-group.spec.js.snap
+++ b/src/shared/common/forms/__tests__/__snapshots__/radio-group.spec.js.snap
@@ -145,7 +145,6 @@ exports[`Form radio group component should render error message for textarea if 
             Some error message
           </span>
           <textarea
-            aria-describedby="ValueAText  ValueAText-error"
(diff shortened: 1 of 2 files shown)


zuzak (commented):
You haven't updated the snapshots for this change, so the tests failed.
They should have run on the precommit hook?

Ideally we would fix the broken reference instead of removing them entirely. Do we need to update our accessibility statement to reflect that we don't have ARIA tags?

It would be good if you could rephrase your commit message so the first line is 50 characters or fewer, so it doesn't wrap in the GitHub UI (and in `git log`):
image

You can do that with the command line with
```
$ git commit --amend -m <<- "EOF"
HOCS-2395 Remove broken ARIA reference

This commit removes the aria-describedby field from the radio group
component as it was broken and for consistency, as most other fields
aren't described by ARIA.
EOF

$ git push origin +HOCS-2395-Remove-broken-aria-reference
```

parth-ho (approved):

PR #584, opened 2021-02-03
HOCS-2444 Add health-checks to Compose; add Readme

zuzak:
This commit improves the Docker Compose experience for Frontend
somewhat, by adding documentation for Docker Compose, and making sure
that Docker brings the containers up in a sensible order (postgres
before info-service, for example; or Localstack before docs). I think
this has always been the intention with the Docker Compose file, as the
`depends_on` properties were already defined -- however, they only
waited for containers to *start*, and didn't wait for the containers to
be *ready* (which is now the new behaviour).

The feature this relies on was removed from Docker 3, but we're not
using any Docker 3 features so it's safe to drop it back down to Docker
2. (We're not using Docker Swarm, you see.) Any docker-compose version
devs are already using will still support this older version of the
syntax.

We add a Readme as this is likely the first part of DECS a new developer
will touch, so this improves the onboarding slightly.

Startup times for Docker Compose might be increased by a minute or so
compared to previously, as we now wait for Localstack to come up safely
before running most of the services. This commit might also mean that
there are more errors than previously -- Compose might complain about
unhealthy containers -- but this is actually just Docker being louder:
these errors will have still existed, but you wouldn't have noticed them
previously.

This commit probably removes the requirement to have separate
"infrastructure" and "services" startup scripts, but I've left them in
so I don't disrupt existing developer workflow.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,156 @@
+# docker-compose
+
+This directory contains configuration for Docker Compose, a quick way to get a
+local HOCS environment on your computer.
+
+To start all services, do:
+```console
+$ docker-compose up
+```
+
+This might take over five minutes to run.
+The first time you run it Compose will download the latest images; this could
+take an additional fifteen minutes, depending on your connection.
+
+You&rsquo;ll also need to configure the database schema;
+[see below for details](#missing-sql-tables).
+
+You can also ask Compose to background the process (detach) once everything is
+up and running:
+```console
+$ docker-compose up -d
+```
+
+To turn the containers off, press CTRL and C at the same time in that terminal
+(SIGINT). Alternatively, run:
+```console
+$ docker-compose down
+```
+
+This will tear down the environment, so you&rsquo;ll have a lengthy wait when
+starting the containers next time. If you&rsquo;re going to need the environment
+again soon, consider instead using the equivalent of standby:
+
+```console
+$ docker-compose pause
+$ docker-compose unpause
+```
+
+To only run the infrastructure containers (like PostgreSQL, AWS and ClamAV),
+or to run the other HOCS services that aren&rsquo;t frontend, do:
+```console
+$ ./scripts/infrastructure.sh
+$ ./scripts/services.sh
+```
+
+You can also run individual services. For example:
+```console
+$ docker-compose up frontend # localhost:8080
+```
+
+If you ask to bring up a service with a dependency, Compose will treat this
+command as an implicit request to bring up the dependencies before bringing
+up the dependant. If you don&rsquo;t want that, use:
+
+```console
+$ docker-compose up --no-deps frontend
+```
+
+If Docker doesn&rsquo;t have the images it needs, it will automatically
+pull them. However, it won&rsquo;t pull existing images. If you want to run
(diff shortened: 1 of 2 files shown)


docker/README.md
```diff
@@ -0,0 +1,156 @@
...
+$ docker-compose up
+```
+
+For more serious problems you might need to tear down the environment with
+`docker-compose down` and start again.
+
+### Missing SQL tables
+
+The info-service doesn&rsquo;t come with its schema, so if you&rsquo;re setting
+up your environment for the first time your info-service you&rsquo;ll need to
```
robertdeniszczyc2, line 119: Think there's a word missing here "your environment for the first time your info-service you'll..."

docker/README.md
```diff
@@ -0,0 +1,156 @@
...
+do them manually.
+
+The info-service schema isn&rsquo;t currently available publicly, and is stored
+in a private repository on GitHub. Request access from an existing team member.
+If you&rsquo;re working in Government but aren&rsquo;t a member of the DECS
+development team, ask someone from the Home Office in the cross-Government Slack
+channel to ask us on your behalf. (We might not be able to give you full access,
+but should be able to send you the basic schemas.)
```
robertdeniszczyc2, line 127: Maybe remove the last line here - we can evaluate on a case by case basis rather than setting expectations too soon.

docker/docker-compose.yml
```diff
@@ -1,9 +1,11 @@
-version: '3.1'
+version: '2.4' # v3 doesn't support depends_on
```
robertdeniszczyc2, line 1: Think it does support it but not in `condition:` form: https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on

docker/README.md
robertdeniszczyc2, line 127:
```suggestion
channel to ask us on your behalf.
```

philbarton (commented):
Whilst you are making changes to docker-compose.yml, please could you change the documents service **SPRING_FLYWAY_SCHEMAS** value to **document** instead of **docs**

docker/docker-compose.yml
zuzak: Yep; you're right. They removed the feature because in Docker Swarm each app is supposed to just retry until it works, instead of quitting if there's a problem, but that's not how HOCS is architected.

docker/README.md
zuzak: Fixed. Looks like I got my tenses confused.
It's not actually "the first time you run your environment", but more "every time you have a new postgres", so I've reworded it.

docker/README.md
zuzak: > Whilst you are making changes to docker-compose.yml

Probably should be a different PR, but I've added a commit to this one anyway.
(Can't put it in the same commit as it wouldn't be atomic.)

philbarton (approved):
Thanks, the document schema was a constant minor irritation!

robertdeniszczyc2 (approved):
Approved

zuzak:
The Drone build is taking 20 minutes to run, but I think that's caused by an earlier commit; nothing in these changes are actually touched by the Drone runner at all.

PR #585, opened 2021-02-04
HOCS-2403-Mock date in tests

agraham19:
CS-2403-Internal Bug: DateInput Component sets maxYear to current date + 100 years breaking tests every year
 
  * mock date in tests so snapshot tests don't break
  * modify component to evaluate date during creation, rather than on import

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2794,7 +2794,7 @@
     },
     "async": {
       "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/async/-/async-1.0.0.tgz",
+      "resolved": "http://registry.npmjs.org/async/-/async-1.0.0.tgz",
       "integrity": "sha1-+PwEyjoTeErenhZBr5hXjPvWR6k="
     },
     "async-each": {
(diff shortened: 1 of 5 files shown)


parth-ho:
Shouldn't there be a snapshot to go with the test changes?

agraham19:
> Shouldn't there be snapshot changes to go with the test changes?

Because I've mocked the date to 2019, it still produces 2119, so the snapshot does not need to change.

parth-ho:
Within the snapshot file there should be a snapshot for each test case where you use `toMatchSnapshot`, since there's a new test case: `should render when maxYear is passed` the snapshot file will also have been updated. If you run `npm test` on this branch you should see that the new test snapshot has been added to the existing snapshot file.

agraham19:
> Within the snapshot file there should be a snapshot for each test case where you use `toMatchSnapshot`, since there's a new test case: `should render when maxYear is passed` the snapshot file will also have been updated. If you run `npm test` on this branch you should see that the new test snapshot has been added to the existing snapshot file.

Ah. Sorry, I've added it now.

parth-ho (approved):
Looks good! The `jest-date-mock` package will defo come in handy for future date related tests, as testing code that handles dates can be a bit tricky.

PR #586, opened 2021-02-09
HOCS-2465: Avoid hideConditions & visibilityConditions conflicts.

parth-ho:
This PR fixes the validation error bug when both hideConditions & visibilityConditions have been defined on a field and in turn causes a conflict. Tests have also been written to cover the different scenarios.


agraham19 (approved):

zuzak:
I think this broke the build; the Docker build step is taking 20 minutes to complete.

PR #587, opened 2021-02-09
HOCS-2282 Remove UNSAFE_ lifecycle methods

robertdeniszczyc2:
React component lifecycle methods prefixed with UNSAFE_ should not be used in production code.
This change removes the unsafe deprecated methods and achieves the same functionality using supported methods.


LiamMacP (approved):
Changes look good - worth pulling `error` and `match` from the props in page-enabled.jsx:16/18/19

robertdeniszczyc2:
@LiamMacP I've made the requested change if you would like another look, thankyou!

LiamMacP:
Looks good to me, ready to merge 🙂

PR #588, opened 2021-02-10
HOCS-2361 Upgrading NodeJS

SeanReedHomeOffice:
This upgrades NodeJS from 8 to 14.15.4, which is the latest stable version to use. 
The 'node:14.15.4-alpine' base imaged is used, which is also used by management-ui.


package.json
```diff
@@ -7,8 +7,8 @@
         "start": "node index.js",
```
parth-ho, line 7: Also needs the flag here
```suggestion
        "start": "node --max-http-header-size 80000 index.js",
```

parth-ho (changes requested):
Looks good, just missing the header size override flag for the start command.

zuzak:
Squash this when merging.

server/lists/adapters/__tests__/case-notes.spec.js
```diff
@@ -28,10 +28,10 @@ describe('Case Notes Adapter', () => {
...
-            { eventTime: '2019-01-05 12:00:01', type: 'STAGE_ALLOCATED_TO_TEAM', userName: 'User A', body: { caseNote: 'A test case note', userUUID: 1, teamUUID: 1, stage: 1 }, timelineItemUUID: '__timelineItemUUID__' },
-            { eventTime: '2019-01-05 12:00:02', type: 'STAGE_CREATED', userName: 'User A', body: { caseNote: 'A test case note', userUUID: 1, teamUUID: 1, stage: 1 }, timelineItemUUID: '__timelineItemUUID__' },
-            { eventTime: '2019-01-05 12:00:03', type: 'STAGE_COMPLETED', userName: 'User A', body: { caseNote: 'A test case note', userUUID: 1, teamUUID: 1, stage: 1 }, timelineItemUUID: '__timelineItemUUID__' },
-            { eventTime: '2019-01-05 12:00:04', type: 'STAGE_RECREATED', userName: 'User A', body: { caseNote: 'A test case note', userUUID: 1, teamUUID: 1, stage: 1 }, timelineItemUUID: '__timelineItemUUID__' },
+            { eventTime: '2019-01-05 12:00:05', type: 'STAGE_ALLOCATED_TO_TEAM', userName: 'User A', body: { caseNote: 'A test case note', userUUID: 1, teamUUID: 1, stage: 1 }, timelineItemUUID: '__timelineItemUUID__' },
```
LiamMacP, line 31: What issue caused these case-note tests to need changing?

SeanReedHomeOffice: Some of the records in the array of case-notes have identical eventTimes and that is the field used for ordering. When upgrading to Node14 these records get ordered in a different way by default. So my fix here is to give unique event times to each row, so the order is defined. See https://homeofficedigital.slack.com/archives/C01GS5FD116/p1612799524139000 for the team chatting about this issue.

LiamMacP: Makes sense and a good solution. Approved.

zuzak (commented):
This PR looks fine, although the linter's failing -- which needs fixing before we can add this branch to master. You should also bring this branch up to date with master. You should also accept Parth's suggestion to pass the header size setting into `npm start`, which is the way production runs the app. At the moment you've only fixed it locally for devs.

Commit hygiene stuff:

Pull requests (mostly), and commits (always) should be atomic; that is, should cover exactly one piece of work apiece. This commit adds whitespace changes, which aren't related to the version bump, so ideally would be committed separately.

Similarly, although the unit test change is related to this NodeJS version bump, it's not clear initially, so an explanation in the Git commit message would be helpful. Your Git commit message at the moment describes _how_ the code changes (fixing unit tests, changing Dockerfile, adding a max-header setting) -- but I know that already (I've read the code) so it's not very useful. It would be better if you used the Git commit message to [explain _why_ you're doing what you're doing](https://chris.beams.io/posts/git-commit/#why-not-how) (in this case, upgrading the Node version, and fixing side-effects). A better commit message might be:

```console
$ git commit --amend -m <<EOF
HOCS-2361 Upgrade NodeJS to latest LTS version

This commit upgrades the version of NodeJS we're using from the
currently unsupported version 8 to the current long-term support
release, version 14.

A change in behaviour in NodeJS in the intervening versions means that
we need to explicitly increase our maximum header size upon startup with
a command-line flag on the interpreter, in order to support our very
large user group headers. We pick 80kB as that was the default in
version 8.

Upgrading to 14 caused one of the unit tests to break because it changes
the way fields are ordered. To ensure a consistent ordering, we ensure
each unit test has a unique value in the sorted field.
EOF
```

This way, a developer chasing down a regression in six months time can look at your commit message and understand what's going on, without resorting to clicking through external sources like Jira tickets and PR comments. (That maintainer could be you.)

If you're wanting to change your commit message you'll probably need to `git rebase -i origin master`, and ask the computer to "edit" the first  commits, and "fixup" the subsequent ones. Alternatively, use the squash feature on GitHub (but rebasing manually would be better).

.drone.yml
```diff
@@ -54,9 +54,9 @@ pipeline:
       image: quay.io/ukhomeofficedigital/sonar-scanner:v3.0.2
       when:
         event: [push, pull_request, tag]
-    
-    
-  
+
```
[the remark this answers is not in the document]
SeanReedHomeOffice: I removed these whitespace changes, they were accidental

package.json
SeanReedHomeOffice: Thanks, I have now done this.

parth-ho (approved):

LiamMacP (approved):

PR #589, opened 2021-02-10
HOCS-2367-WCS_Option for bulk transfer of cases

agraham19:
* add handler to differentiate between submit buttons
* add move team option
* invoke casework api to move cases between teams

### DIFF
--- a/server/middleware/workstack.js
+++ b/server/middleware/workstack.js
@@ -52,11 +52,34 @@ async function getTeamMembers(req, res, next) {
     }
 }
 
+async function getMoveTeamOptions(req, res, next) {
+    try {
+        const response = await req.listService.fetch('MOVE_TEAM_OPTIONS', req.params);
+        res.locals.workstack.moveTeamOptions = response;
+        next();
+    } catch (error) {
+        next(error);
+    }
+}
+
 function workstackApiResponseMiddleware(req, res) {
     res.json(res.locals.workstack);
 }
 
-const allocateUser = async (req, res, [endpoint, body, headers]) => {
+const sendMoveTeamRequest = async (req, res, [endpoint, body, headers]) => {
+    const logger = getLogger(req.requestId);
+    try {
+        await caseworkService.put(endpoint, body, headers);
+        return;
+    } catch (error) {
+        logger.error('MOVE_TEAM_FAILED', { endpoint, body, status: error.response.status });
+        res.locals.notification = 'Failed to transfer all cases to team';
+        return;
+    }
+};
+
+
+const sendAllocateUserRequest = async (req, res, [endpoint, body, headers]) => {
     const logger = getLogger(req.requestId);
     try {
         await caseworkService.put(endpoint, body, headers);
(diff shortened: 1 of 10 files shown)


server/middleware/workstack.js
```diff
@@ -52,11 +52,21 @@ async function getTeamMembers(req, res, next) {
...
+        next(error);
+    }
+}
+
 function workstackApiResponseMiddleware(req, res) {
     res.json(res.locals.workstack);
 }
 
-const allocateUser = async (req, res, [endpoint, body, headers]) => {
+const putToCaseworkService = async (req, res, [endpoint, body, headers]) => {
```
robertdeniszczyc2, line 69: This isn't strictly descriptive of the behaviour described in the method. If the functionality fails the try and is caught, it will log an error about allocation failure, where the method name suggests it's responsible for more agnostic 'putting'

server/middleware/workstack.js
```diff
@@ -79,12 +103,35 @@ async function allocateToTeam(req, res, next) {
...
+    }
+
+    if (selected_cases.length > 0 && selected_team) {
+        const requests = selected_cases
+            .map(selected => selected.split(':'))
+            .map(([caseId, stageId]) => [`/case/${caseId}/stage/${stageId}/team`, { teamUUID: selected_team }, {
+                headers: User.createHeaders(req.user)
+            }])
+            .map(async options => await putToCaseworkService(req, res, options));
+        await Promise.all(requests);
```
robertdeniszczyc2, line 128: Just to check if this fails where is the error caught? Tried to follow where it would bubble up to but couldn't see immediately where it would be caught.

server/middleware/workstack.js
```diff
@@ -79,12 +103,35 @@ async function allocateToTeam(req, res, next) {
...
                 headers: User.createHeaders(req.user)
             }])
-            .map(async options => await allocateUser(req, res, options));
+            .map(async options => await putToCaseworkService(req, res, options));
         await Promise.all(requests);
     }
     next();
 }
 
+async function moveTeam(req, res, next) {
```
robertdeniszczyc2, line 112: I can't see any tests for this method, may need some adding

server/middleware/workstack.js
agraham19: I could make the error message more generic, this would make the error less informative, but it would still have the endpoint in the message. Or do you think it would be better to put the name back and create a dedicated method for move team?

robertdeniszczyc2: I think it might be better to split them out and have the method name more descriptive of the behaviour it is performing. If it was like the previous allocateUser and another for move team, it gives us more flexibility to do more stuff in future for those actions separately.

The other option could be to pass in the error and notification messages as another parameter into the method, but this could make the method more difficult to understand its purpose.

server/middleware/workstack.js
agraham19: The tests for "handleWorkstackSubmit" cover this method when the, but I've renamed this to "workstack submit case update (allocate/move team)" as it covers the whole path.

server/middleware/workstack.js
agraham19: I have split these

server/middleware/workstack.js
robertdeniszczyc2: Hmm that's interesting, yeah feel free to raise a ticket, that's a good spot!

robertdeniszczyc2 (approved):

PR #590, opened 2021-02-12
HOCS-2477-Cannot-Filter-By-Correspondent-Or-Ref

agraham19:
* where a column entry is an object, stringify its values for filtering

### DIFF
--- a/src/shared/common/components/workstack.jsx
+++ b/src/shared/common/components/workstack.jsx
@@ -69,7 +69,12 @@ class WorkstackAllocate extends Component {
     }
 
     doesFilterMatchData(filter, row, column) {
-        const value = this.getCellValue(row, column);
+        let value = this.getCellValue(row, column);
+
+        if (value && Object.keys(value).length !== 0 && value.constructor === Object) {
+            value = Object.values(value).join('');
+        }
+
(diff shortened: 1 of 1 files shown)


src/shared/common/components/workstack.jsx
```diff
@@ -69,7 +69,12 @@ class WorkstackAllocate extends Component {
     }
 
     doesFilterMatchData(filter, row, column) {
-        const value = this.getCellValue(row, column);
+        let value = this.getCellValue(row, column);
+
+        if (typeof value === 'object' && value != null) {
```
robertdeniszczyc2, line 74: To make this a concrete check for an object use:
```suggestion
        if (value && Object.keys(value).length !== 0 && value.constructor === Object) {
```
as the current implementation would fail if for some reason value was a RegEx. I think in this context this is highly unlikely however.

See https://stackoverflow.com/a/32108184 for more info

robertdeniszczyc2 (commented):
Added a minor comment about the object check. I think the scenario I've presented is highly unlikely, but might as well make it as resilient as possible.

src/shared/common/components/workstack.jsx
agraham19: I have addressed this.

robertdeniszczyc2 (approved):
Changes approved

PR #592, opened 2021-02-22
HOCS-2473: Allow UKVI to mark overdue contribution as received

parth-ho:
This PR changes the contributionDueDate validator to allow dates to be 6 months in the past.

PR #596, opened 2021-03-01
HOCS-2473: Add isValidWithinDate validator to contribution-fulfillment.js

parth-ho:
The change in validation for the contribution-fulfillment form was missed in the last [PR](https://github.com/UKHomeOffice/hocs-frontend/pull/592/files). This PR adds the validator to this form.

### DIFF
--- a/server/services/forms/schemas/contribution-fulfillment.js
+++ b/server/services/forms/schemas/contribution-fulfillment.js
@@ -49,7 +49,7 @@ module.exports = async options => {
             Component('date', 'contributionDueDate')
                 .withValidator('required')
                 .withValidator('isValidDate')
-                .withValidator('isAfterToday')
+                .withValidator('isValidWithinDate')
                 .withProp('label', 'Contribution due date')
                 .build()
         )


LiamMacP (approved):

zuzak:
It's not immediately clear from the commit message what this change does. Ideally you'd [explain _why_](https://chris.beams.io/posts/git-commit/#why-not-how): I know you're adding the `isValidWithinDate` validator to `contribution-fullfillment.js`, because that's what the rest of the commit says. What I don't know, from looking at the commit in isolation, is why that's important. If there's a bug as a result of this change, the maintainer would need to click through to Jira to try and understand this change.

Can you change the commit to explain why you're changing the validator?

parth-ho:
@zuzak I'm not opposed to this, but I think there should be discussion team discussion around it so we are all on the same page as to what is expected.

IMO I think the "why" (or business context or purpose of a commit) can be sometimes difficult to squeeze in a single commit, this is why we have the JIRA number prefixed, where the "why" is much better placed; looking up a ticket in JIRA quite easy. A jira ticket can have a lot of information (i.e. comments, links, dependencies, history, status, release info etc), and I think it's wise or practical to try and squeeze all of this info into every commit message, because at the time of the commit we can't possibly know the bit of information which could be useful if/when there is a bug. As a personal preference I just like to know what has been done on a commit message (clear & straight to the point), then if I think I need more info I will look up the ticket. I'be curious as to how often we look at commit messages and what value we expect from them.

Maybe for a "proper" open source/community project (which DECS is not) having as much info as possible in a commit makes sense, but as this is business driven and there can be many complicated reasons for doing something, and given we use JIRA, I don't think blocking PR's to do this is worth it IMHO.

zuzak:
Oh, yes, it wasn't a blocking comment (I'd've used the "request changes" if it were) -- feel free to merge.

My view is that commits should stand alone -- they don't need to be comprehensive by any means, but just enough that a developer who only vaguely knows this codebase can get a quick understanding of the change without having to refer elsewhere. Here, I'd have a commit message like "HOCS-2473 Allow contribution due date to be in past".

PR #599, opened 2021-03-09
HOCS-2560 Added Case Creator healthcheck

robertdeniszczyc2:


server/clients/index.js
```diff
@@ -40,4 +40,4 @@ module.exports = {
     infoService,
     templatesService,
     documentService
-};
\ No newline at end of file
+};
```
zuzak, line 43: Ideally these changes would be in a separate commit. (And you'd rebase away the intermediate commit that does nothing, I guess?)

philbarton (commented):
We’ll need a way for the docker-compose to pull the images for the correct epic branch. Currently I think it pulls images built from master

server/clients/index.js
robertdeniszczyc2: I don't actually know where these changes are from, I'd rebased away the commit that does nothing and although the files appear in a `git show --word-diff` there is no change shown. IntelliJ also doesn't track a change here with its Git Blame annotations.

zuzak: It's not too important; it's not worth blocking a merge over.

PR #610, opened 2021-03-26
HOCS-2623 Update docker-compose for case creator

robertdeniszczyc2:
Adds required envvars for case creator and updates docker compose to use tags on images rather than pulling latest by default. Allows for running services depending on current workstream.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -128,8 +128,7 @@ services:
   frontend:
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8080/actuator/health"]
-    image: quay.io/ukhomeofficedigital/hocs-frontend
-    # image: hocs-frontend-test
+    image: quay.io/ukhomeofficedigital/hocs-frontend:${FRONTEND_TAG:-latest}
     ports:
     - 8080:8080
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yml
```diff
@@ -128,7 +128,7 @@ services:
   frontend:
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8080/actuator/health"]
-    image: quay.io/ukhomeofficedigital/hocs-frontend
+    image: quay.io/ukhomeofficedigital/hocs-frontend:${FRONTEND_TAG}
```
zuzak, line 131: You can use shell expansion here!
That way you'll only need to put non-mainline versions into your `.env`.
```suggestion
    image: quay.io/ukhomeofficedigital/hocs-frontend:${FRONTEND_TAG:-latest}
```

robertdeniszczyc2: Ohhh I didn't realise that, that's cool. However by doing that how do we manage both sets of envvars in the `.env` file, just comment out the ones we don't need at the time or something?

philbarton (approved):
Looks good to me.

PR #617, opened 2021-04-12
Hocs 2527 additional config case creator

philbarton:
Add missing config to get the case creator working in docker.

Also :
- Fixed postgres error
- Added a sleep, to aws_cli. Without the sleep, it gets run multiple times.
- Set images to main rather than latest

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -162,7 +162,7 @@ services:
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8080/actuator/health"]
       start_period: 240s
-    image: quay.io/ukhomeofficedigital/hocs-casework:${CASEWORK_TAG-branch-main}
+    image: quay.io/ukhomeofficedigital/hocs-casework:${CASEWORK_TAG:-branch-main}
     ports:
     - 8082:8080
     networks:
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yml
```diff
@@ -159,8 +162,7 @@ services:
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8080/actuator/health"]
       start_period: 240s
-    image: quay.io/ukhomeofficedigital/hocs-casework
-    #image: hocs-casework-test
+    image: quay.io/ukhomeofficedigital/hocs-casework:${CASEWORK_TAG-branch-main}
```
LiamMacP, line 165: Does the following need the colon like the fronend above?

zuzak:
Once you've got to the point that this is ready to be merged, close this PR and reopen it against [ukhomeoffice/hocs](https://github.com/UKHomeOffice/hocs) instead.

I'm not sure what the future of hocs-frontend's docker-compose should be; maybe we should just remove it?

docker/docker-compose.yml
philbarton: It works with both syntax's!! I've change them all to ${X_TAG:-branch-main} for consistancy

robertdeniszczyc2 (approved):
Looks good to me

PR #621, opened 2021-04-26
HOCS 2709 add without mp

philbarton:
When creating a complaint, selecting wether the complaint is an MP is irrelevant.
This change modifies the correspondent creation to allow the skipping of the MP step.
This is achieved by passing 'addUrlPath' = 'addNoMp' when defining the entity-list field.

### DIFF
--- a/server/services/forms/schemas/add-correspondent-details.js
+++ b/server/services/forms/schemas/add-correspondent-details.js
@@ -1,6 +1,14 @@
 const Form = require('../form-builder');
 const { Component, Choice } = require('../component-builder');
 
+function buildUrl(options) {
+    if (options.action === 'addNoMp') {
+        return `/case/${options.caseId}/stage/${options.stageId}`;
+    } else {
+        return `/case/${options.caseId}/stage/${options.stageId}/entity/correspondent/add`;
+    }
+}
+
 module.exports = options => Form()
     .withTitle('Record Correspondent Details')
     .withField(
@@ -64,7 +72,7 @@ module.exports = options => Form()
     .withSecondaryAction(
         Component('backlink')
             .withProp('label', 'Back')
-            .withProp('action', `/case/${options.caseId}/stage/${options.stageId}/entity/correspondent/add`)
+            .withProp('action', buildUrl(options))
             .build()
     )
     .build();
--- a/server/services/forms/schemas/add-correspondent-details.js
+++ b/server/services/forms/schemas/add-correspondent-details.js
@@ -2,11 +2,11 @@ const Form = require('../form-builder');
 const { Component, Choice } = require('../component-builder');
 
 function buildUrl(options) {
+    const url = `/case/${options.caseId}/stage/${options.stageId}`;
     if (options.action === 'addNoMp') {
-        return `/case/${options.caseId}/stage/${options.stageId}`;
-    } else {
-        return `/case/${options.caseId}/stage/${options.stageId}/entity/correspondent/add`;
+        return url;
     }
+    return url + '/entity/correspondent/add';
 }
 
 module.exports = options => Form()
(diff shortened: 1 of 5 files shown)


src/shared/common/forms/composite/entity-list.jsx
```diff
@@ -96,6 +97,7 @@ class EntityList extends Component {
                             })}
                             <br />
                             {hasAddLink && <Link to={`/case/${page.params.caseId}/stage/${page.params.stageId}/entity/${entity}/add?hideSidebar=${hideSidebar}`} className="govuk-body govuk-link">Add a {entity}</Link>}
+                            {hasAddLinkNoMp && <Link to={`/case/${page.params.caseId}/stage/${page.params.stageId}/entity/${entity}/addNoMp?hideSidebar=${hideSidebar}`} className="govuk-body govuk-link">Add a {entity}</Link>}
```
LiamMacP, line 100: I wonder if you genericise this like the `sumo-list`. So have a primaryLink item instead that changes the where the action should point to? 

`{hasAddLinkNoMp && <Link to={`/case/${page.params.caseId}/stage/${page.params.stageId}/entity/${entity}/addNoMp?hideSidebar=${hideSidebar}`} className="govuk-body govuk-link">Add a {entity}</Link>}` 
would then become something similar to 
`{primaryLink && <Link to={`/case/${page.params.caseId}/stage/${page.params.stageId}/entity/${entity}/{primaryLink.action}?hideSidebar=${hideSidebar}`} className="govuk-body govuk-link">Add a {entity}</Link>}`

Would be a little work but it would be a copy from `somu-list` pretty much.

src/shared/common/forms/composite/entity-list.jsx
```diff
@@ -112,6 +114,7 @@ EntityList.propTypes = {
     disabled: PropTypes.bool,
     error: PropTypes.string,
     entity: PropTypes.string.isRequired,
+    hasAddLinkNoMp: PropTypes.bool,
```
LiamMacP, line 117: If you go down the route above this would no longer be needed. I'm not particularly happy with bloating the generic components with specifics if possible.

server/services/forms/schemas/add-correspondent-details.js
```diff
@@ -1,6 +1,14 @@
 const Form = require('../form-builder');
 const { Component, Choice } = require('../component-builder');
 
+function buildUrl(options) {
+    if (options.action === 'addNoMp') {
+        return `/case/${options.caseId}/stage/${options.stageId}`;
+    } else {
+        return `/case/${options.caseId}/stage/${options.stageId}/entity/correspondent/add`;
+    }
+}
```
robertdeniszczyc2, line 10: Slightly cleaner like:
```suggestion
function buildUrl(options) {
    let url = `/case/${options.caseId}/stage/${options.stageId}`;

    if (options.action === 'addNoMp') {
        return url;
    }

    return url + `/entity/correspondent/add`;
}
```

Could also add tests too

src/shared/common/forms/composite/entity-list.jsx
robertdeniszczyc2, line 100: Can both of these ever be shown or is it one or the other?

As far as I can see here the difference in URL is the last bit at the end, so if only one is shown you could set that value as a variable with some logic and then only be having the Link component here once avoiding the duplication.

robertdeniszczyc2 (changes requested):

server/services/forms/schemas/add-correspondent-details.js
zuzak: That `let` can be a `const`.

robertdeniszczyc2: Indeed it can, fixed :)

robertdeniszczyc2 (approved):

robertdeniszczyc2 (changes requested):
Build is failing, test needs updating

robertdeniszczyc2 (approved):
Approved pending build passes

PR #622, opened 2021-04-28
HOCS-2674: change teams type

LiamMacP:
Currently, the teams endpoint is static and pulls the fresh list from
info on every pod start. As we now allow for dynamic teams to be
added through MUI this can cause discrepancies between data. This
change enforces that the team is pulled on every request.

### DIFF
--- a/server/lists/index.js
+++ b/server/lists/index.js
@@ -30,7 +30,6 @@ module.exports = {
         S_TEAMS: {
             client: 'INFO',
             endpoint: '/team',
-            type: listService.types.STATIC,
             adapter: statics.teamsAdapter
         },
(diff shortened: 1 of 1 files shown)


zuzak:
Can we explicitly mark it `DYNAMIC` instead of implicitly marking it "not `STATIC`"?

LiamMacP:
> Can we explicitly mark it `DYNAMIC` instead of implicitly marking it "not `STATIC`"?

This has been changed, I have changed all of the team endpoint calls that strictly deal with returning teams to DYNAMIC

Paulcr-ho (approved):

PR #625, opened 2021-04-29
HOCS-2674: static cache scheduling

LiamMacP:
Currently our implementation of caching means removing it all together is unfeasible. This change proposes to add a static cache scheduler cronjob that periodically flushes all of the static data held on the system.

Too allow for this, the following changes have been implemented: 
- Added `node-cron` as a dependency to allow for cron jobs. Theoretically, we could have used `setInterval` for this, but having this cronjob happen as a predictable time allows for better transparency with users and reduces potential issues going forward.
- The schedule for this was added as a config variable (`STATIC_DATA_REFRESH_SCHEDULE`). The default though for this is to run at 0, 15, 30, 45 of the hour while the pod is up. We could further restrict this to working hours, if the requirement is there?
- This new cron job is initialised at the same time the static data is initialised, this seemed like the best place as at that point we had that data. This reduces the risk of a cyclic method call issue.
- Currently, there is still going to be a chance that a team and assigned a case within 15 minutes before the static data cache call. This is through an edge case with a pod coming up and loading a new team while an old pod doesn't have the team. To handle this, there is a optional parameter `flushIfNull` that defaults to false. Changing this to true for the teams allows us to make sure that if a team is not found we can load the fresh data.
- To allow for the testing of this new scheduler, a subset of the sonin test pack is used - `@sinonjs/fake-timers` which allows us to handle the mocking of the system clock and ensure that the scheduler fires at the correct time.

** There is still a potential issue with team renames not filtering through till the the next cache restart. This is out of scope of this fix. Something we could look at is [etags](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag), which would allow us to check a not modified header. Our current system does not allow for this though so would be a potential solution going forward.


LiamMacP:
A further improvement to this could be to introduce a `PARTIALLY_STATIC` type and only update that (which would be stuff that can be changed on the fly in MUI). But that could always be part of a further fix. Any opinions?

robertdeniszczyc2:
Think come to `PARTIALLY_STATIC` separately... Static to me sounds boolean, it's either static or it's dynamic, and something that is static but only partially when x/y/z feels a bit odd.

robertdeniszczyc2 (approved):

PR #626, opened 2021-04-30
Hocs 2787 text limit

philbarton:
The problem identified is that, if a camunda flow doesn't make explicit saves, camunda temporarily saves values between steps. When saving, the maximum length is 4000 chars. If this is exceeded then we get the unhelpful, "Something went wrong"

Implementation :
text-areas now have a countdown of characters, with a default maximum of 4000. 
text input has had a maxlen applied with a default maximum of 4000. For text input there is no countdown. 

For both, the default can be changed by passing the property e.g. "limit="300" from the hocs-data definition.


src/shared/common/forms/text-area.jsx
```diff
@@ -42,11 +43,21 @@ class TextArea extends Component {
                     disabled={disabled}
                     rows={rows}
                     onChange={e => this._onChange(e)}
+                    value={this.state.value}
                     defaultValue={this.state.value}
                 />
+                { ! disabled &&
```
robertdeniszczyc2, line 49:
```suggestion
                {!disabled &&
```

robertdeniszczyc2 (approved):

PR #627, opened 2021-05-04
HOCS-2975: update case data CaseworkTeamUUID

LiamMacP:
Currently when we move a WCS case it does not update the case
data. This change hits the workflow service update data endpoint
which triggers an internal camunda update and then updates the
casedata as well.


server/middleware/workstack.js
```diff
@@ -133,12 +145,23 @@ async function moveTeam(req, res, next) {
...
-            }])
-            .map(async options => await sendMoveTeamRequest(req, res, options));
-        await Promise.all(requests);
+            .map(selected => selected.split(':'));
+
+        const updateCaseDataRequests =
+            requests
+                .map(([caseId, stageId]) => [`/case/${caseId}/stage/${stageId}/CaseworkTeamUUID`, { value: selected_team }, {
+                    headers: User.createHeaders(req.user)
+                }])
```
zuzak, line 154: From a code-style point of view I'd have either everything on one line or everything on different lines (preferred, as it plays nicer with `diff`) -- but not both:
```suggestion
                .map(([caseId, stageId]) => [
                  `/case/${caseId}/stage/${stageId}/CaseworkTeamUUID`,
                  { value: selected_team },
                  { headers: User.createHeaders(req.user) }
                ])
```

server/middleware/workstack.js
```diff
@@ -66,14 +66,26 @@ function workstackApiResponseMiddleware(req, res) {
     res.json(res.locals.workstack);
 }
 
+const updateCaseDataMoveTeamRequest = async (req, res, [endpoint, body, headers]) => {
+    const logger = getLogger(req.requestId);
+    try {
+        await workflowService.put(endpoint, body, headers);
```
zuzak, line 72: This seems to be a bit of a waste of an `async`/`await` -- this is essentially blocking anyway? (I'm never sure how async/await works).

server/middleware/workstack.js
```diff
@@ -66,14 +66,26 @@ function workstackApiResponseMiddleware(req, res) {
...
+        return;
```
zuzak, line 73: These `return undefined` doesn't serve any purpose as it's at the end of the function.

server/middleware/workstack.js
```diff
@@ -66,14 +66,26 @@ function workstackApiResponseMiddleware(req, res) {
...
+    } catch (error) {
+        logger.error('MOVE_TEAM_FAILED', { endpoint, body, status: error.response.status });
+        res.locals.notification = 'Failed to update all cases data with new team.';
```
zuzak, line 76: I think this copy needs work. If it's used as part of the notification element, we'll only have 1000ms to read and understand what's going on, so it needs to be brief.

(I would rather it threw a real 500 page, but we've got to work with what we've got, I guess)

server/middleware/workstack.js
LiamMacP: True will remove.

server/middleware/workstack.js
LiamMacP: Yeah we don't actually use these `res.locals.notification` anywhere. I'm happy to keep it in for now and open a ticket to review. As we should ideally be showing this error with a 500 page as you say.

server/middleware/workstack.js
LiamMacP: Agreed, will change it 👍🏼

server/middleware/workstack.js
LiamMacP: I think it's because of the way we handle middleware. This is currently used in a rather funky way but could be used elsewhere, where we need to wait for it to finish before progressing further middleware.

server/middleware/workstack.js
LiamMacP: This has been rectified.

server/middleware/workstack.js
LiamMacP: This has been rectified.

server/middleware/workstack.js
LiamMacP: We need this to handle the promise rejection on failure. I'll put this down to a limitation with our implementation. I do agree though, we should be handling this on a higher level.

server/middleware/workstack.js
LiamMacP: HOCS-2917 has been created for this, resolving for now.

PR #634, opened 2021-05-12
HOCS-2710 Form validation

robertdeniszczyc2:
Modifies validation mechanism to allow for form level validators


server/middleware/validation.js
```diff
@@ -175,15 +188,26 @@ function validationMiddleware(req, res, next) {
...
+            const fieldValidationErrors = schema.fields
                 .filter(field => field.type !== 'display')
                 .reduce((result, field) => {
                     validateField(field, data, result, validationSuppressor);
                     return result;
                 }, {});
+
+
+            let formValidationErrors = {};
+
```
SeanReedHomeOffice, line 202: I am not sure this implementation would handle the current need on the Multiple Contributions screen for form level validation. On the MCs add/edit form one date in the form needs to be earlier another date in the same form. In this case we need both field level validation and form level validation on the same fields.

If we perform form level validation only if there are no field level validation errors then we have only one error message per field and I think its fixed.

robertdeniszczyc2: In that instance we would need to go with the latter currently, as the components and error handling architecture isn't built for handling multiple failures per field. Also there would be design issues to contend with, there's nothing in the Design System as far as I know to handle this, particularly around the summary messages linking to fields with errors.

SeanReedHomeOffice: Perhaps the the confluence page you created could have a note about this aspect.

SeanReedHomeOffice (approved):
LGTM

PR #635, opened 2021-05-12
HOCS-2992 Update dependancies

robertdeniszczyc2:


LiamMacP (approved):
Out of interest do we know how we use this? Do you think we should regression test this? Although it's a patch version is there anything that would break backwards compatability?

robertdeniszczyc2:
@LiamMacP It's used for the minification, see webpack.config.js. Always good for a smoke test to check the app runs (I've tried it locally, seems OK).

If we update webpack to v5 we can remove the plugin: https://webpack.js.org/plugins/terser-webpack-plugin/#getting-started

PR #637, opened 2021-05-14
HOCS-2980: Statics lists for approval teams & users

parth-ho:


LiamMacP:
Wow that's alot of static lists, as part of the caching scheduler all these will be refreshed every 15 minutes. Probably want to revisit the caching to only refresh specific ones as these are hardly ever going to change.

abarker66:
Since the lists identify themselves as static should they be refreshed at all? Could they be ignored by the caching scheduler?

LiamMacP:
> Since the lists identify themselves as static should they be refreshed at all? Could they be ignored by the caching scheduler?

It was raised as a potential concern in the following PR: 
https://github.com/UKHomeOffice/hocs-frontend/pull/625

We would ideally have another type that use for a things that should be scheduled for a reload. I will create a ticket for that though...

zuzak (commented):
Is there a better way to code this rather than twenty identical bits of code?
```js
foiEndpoints.map((x) => {
  return {
    client: 'INFO',
    endpoint: `/teams?unit=${x}`,
    type: listService.types.STATIC,
    adapter: usersAdapter
  }
})

Although I guess it'd be better to make the computer know that to get any given unit's team they can hit `/teams?unit=x`. Or simply by taking the `/teams` output and filtering

agraham19 (approved):

PR #638, opened 2021-05-17
HOCS-2977 - Fix rendering of checkboxes in MappedDisplay component.

philbarton:
Change the MappedDisplay component so that it renders checkboxes differently.

Instead of:    "MappedDisplay field: Send to QA"
Render as:    "Send to QA: <disabled checked checkbox>

### DIFF
--- a/src/shared/common/forms/mapped-display.jsx
+++ b/src/shared/common/forms/mapped-display.jsx
@@ -17,7 +17,7 @@ class MappedDisplay extends Component {
         const mappings = new Map(choices.map(choice => [choice.value, choice.label]));
         if (component === 'checkbox') {
             return (
-                <span className='govuk-body full-width'><strong>{choices[0].label}: </strong><input className="bigger" type="checkbox" disabled="disabled" checked="checked"/></span>
+                <span className='govuk-body full-width'><strong>{choices[0].label}: </strong>Yes</span>
             );
         } else {
             return (
--- a/src/shared/common/forms/mapped-display.jsx
+++ b/src/shared/common/forms/mapped-display.jsx
@@ -10,24 +10,33 @@ class MappedDisplay extends Component {
     render() {
         const {
             choices,
+            component,
             label,
             value
         } = this.props;
         const mappings = new Map(choices.map(choice => [choice.value, choice.label]));
-        return (
-            <span className='govuk-body full-width'><strong>{label}: </strong>{mappings.get(value) ? mappings.get(value) : value}</span>
-        );
+        if (component === 'checkbox') {
+            return (
+                <span className='govuk-body full-width'><strong>{choices[0].label}: </strong><input className="bigger" type="checkbox" disabled="disabled" checked="checked"/></span>
+            );
+        } else {
+            return (
+                <span className='govuk-body full-width'><strong>{label}: </strong>{mappings.get(value) ? mappings.get(value) : value}</span>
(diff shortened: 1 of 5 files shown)


src/shared/common/forms/mapped-display.jsx
```diff
@@ -10,26 +10,35 @@ class MappedDisplay extends Component {
...
             label,
             value
         } = this.props;
         const mappings = new Map(choices.map(choice => [choice.value, choice.label]));
-        return (
-            <span className='govuk-body full-width'><strong>{label}: </strong>{mappings.get(value) ? mappings.get(value) : value}</span>
-        );
+        if (component === 'checkbox') {
+            return (
+                <span className='govuk-body full-width'><strong>{choices[0].label}: </strong><input className="bigger" type="checkbox" disabled="disabled" checked="checked"/></span>
```
zuzak, line 20: I know nothing of this ticket, but generally things like 
```html
<strong>Foo:</strong> bar
```
are better written using the `dd` and `dt` elements.

robertdeniszczyc2, line 20: Two questions on this here:

- Are we sure the label field will always exist in the **first** element in the choices variable? Conscious of magic numbers.
- Has this approach been run past designer/assessed for accessibility? A disabled input field used solely for presentational purpose has a bit of a smell, potential that screen readers could completely ignore that an element is even there: https://a11y-101.com/development/aria-disabled

robertdeniszczyc2 (changes requested):

LiamMacP:
Do you know how this works with hidden fields too? Multiple contributions adds a hidden field (to help migration) but they show as `MappedDisplay Field` too.

philbarton:
> Do you know how this works with hidden fields too? Multiple contributions adds a hidden field (to help migration) but they show as `MappedDisplay Field` too. Although it's not linked to checkboxes might be the same fix?

@LiamMacP The use of hidden fields will be slightly different, but in the same area. I could look at HOCS-2989 as well after this ticket.

src/shared/common/forms/mapped-display.jsx
philbarton: The label field is there in all of the existing checkbox choices definitions. It's all magic!

The problem with the way checkboxes are implemented on DECS is that they have all sorts of values. See note on the ticket. 
The guidance didn't help!

I could set it to the text "Selected"?

I'll have a chat with Ash.

LiamMacP:
> > Do you know how this works with hidden fields too? Multiple contributions adds a hidden field (to help migration) but they show as `MappedDisplay Field` too. Although it's not linked to checkboxes might be the same fix?
> 
> @LiamMacP The use of hidden fields will be slightly different, but in the same area. I could look at HOCS-2989 as well after this ticket.

No worries if you haven't gone the time - just thought as you were in the area

src/shared/common/forms/mapped-display.jsx
philbarton: strong used in other components. I'd rather stay with the standard.

philbarton: @robertdeniszczyc2 I spoke to Ash. He has suggested the use of 'Yes' instead of a checkbox.

robertdeniszczyc2 (approved):
Thanks for resolving comments, good from me 👍

PR #647, opened 2021-05-27
Hocs 3077 fix frontend user issue

agraham19:
HOCS-3077 fix frontend user issue
    
Currently the frontend pulls a list of users for each approal team at startup.
When this list is pulled, there is a comparison operator to check if each user
is the requesting user. At startup, however, there is no user, as it's not
triggered by a requested. This crashes the frontend.
   
* Only check the user matches the pulled user if one exists

### DIFF
--- a/server/lists/adapters/users.js
+++ b/server/lists/adapters/users.js
@@ -3,10 +3,11 @@ const byLabel = (a, b) => a.label.toUpperCase().localeCompare(b.label.toUpperCas
 module.exports = async (data, { user, logger }) => {
     logger.debug('REQUEST_USERS', { users: data.length });
     return data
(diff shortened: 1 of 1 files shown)


zuzak:
Please amend your commit message to have a body that explains what's going on. "Fix issue" is too nebulous.
Even if I consult Jira, I don't know why the frontend wasn't working, or how adding `user &&` fixes it.

agraham19:
> Please amend your commit message to have a body that explains what's going on. "Fix issue" is too nebulous.
> Even if I consult Jira, I don't know why the frontend wasn't working, or how adding `user &&` fixes it.

I have added some more detail

SeanReedHomeOffice (approved):
LGTM

PR #652, opened 2021-06-02
HOCS-2847 : COMP search results

MarkWrightBjss:
Server side passes through the whole primary correspondent (full name and postcode).
Client side, new data adapter to pull data from primary correspondent by key.
Merge to epic.


src/shared/common/components/__tests__/workstack.spec.jsx
```diff
@@ -63,6 +69,8 @@ describe('Workstack component', () => {
         columns: [
             { displayName: 'Reference', dataAdapter: null, renderer: 'caseLink', dataValueKey: 'caseReference', isFilterable: true },
             { displayName: 'Full Name', dataAdapter: null, renderer: null, dataValueKey: 'fullName', isFilterable: true },
+            { displayName: 'Fullname', dataAdapter: 'primaryCorrespondent', renderer: null, dataValueKey: 'fullname', isFilterable: true },
```
philbarton, line 72: Slightly confused by this. "Full Name" is the column header. Do we need Fullname ?

philbarton (approved):
Looks good, just one question in the coments. Also, the description in the ticket needs updating to reflect actual Column headers.

PR #653, opened 2021-06-03
HOCS-3095 - Replace empty strings with 'undefined' for search

philbarton:
The search was failing if no DOB was set.

The two new fields ComplainantDOB and ComplainantHORef could have empty strings in them, which were being passed to as search params.

Changing them to 'undefined', if empty,  stops them from being used in the search.

Local testing proves the fix.

### DIFF
--- a/server/middleware/__tests__/searchHandler.spec.js
+++ b/server/middleware/__tests__/searchHandler.spec.js
@@ -36,7 +36,9 @@ describe('handleSearch', () => {
                     'caseStatus': 'active',
                     'CampaignType': 'Test Campaign 123',
                     'MinSignOffTeam': 'Test Min Sign Off Team',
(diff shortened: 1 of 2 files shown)


MarkWrightBjss (changes requested):
With this change, I'm still able to re-create the issue on my local. Was it explicitly a frontend change, or was there a supporting change in Search?

philbarton:
> With this change, I'm still able to re-create the issue on my local. Was it explicitly a frontend change, or was there a supporting change in Search?

It was just a change to the frontend. I'll test again locally! I do clear down the indexes and DB before testing though!

philbarton:
I've just tested this again locally in a clean environment and it doesn't work.
I must have got mixed up during testing! 
Canceling the PR

PR #654, opened 2021-06-04
HOCS-2763: remove OfficialEngagement from search

LiamMacP:
We have removed the OfficialEngagement from the search for MTS
cases and MTS cases going forward. As a result the business no
longer need to be able to search based off this and this can be
removed from passed data.

### DIFF
--- a/server/services/forms/schemas/search.js
+++ b/server/services/forms/schemas/search.js
@@ -16,7 +16,6 @@ const search = async (options = {}) => {
     let fields = [];
     const systemConfiguration = await listService.fetch('S_SYSTEM_CONFIGURATION');
     systemConfiguration.profiles.map(profile => {
-        console.log(profile);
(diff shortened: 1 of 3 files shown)


server/services/forms/schemas/search.js
```diff
@@ -16,6 +16,7 @@ const search = async (options = {}) => {
     let fields = [];
     const systemConfiguration = await listService.fetch('S_SYSTEM_CONFIGURATION');
     systemConfiguration.profiles.map(profile => {
+        console.log(profile);
```
zuzak, line 19: Please don't echo unstructured data to console
```suggestion
```

LiamMacP: Was for local use, mis-commit. Removed.

MarkWrightBjss (approved):

PR #655, opened 2021-06-07
HOCS-2583 request pit extension

agraham19:
* Add new FOI sidebar screen to allow PIT exemptions to be applied to FOI cases


src/shared/common/components/foi-actions.jsx
```diff
@@ -0,0 +1,28 @@
+import React, { Fragment, useContext } from 'react';
```
SeanReedHomeOffice, line 1: I would prefer that tab to say 'Extensions' instead of 'Actions' and display a list of the extension already applied. All the other tabs refer to a type of data that the tab displays. But I know you are following the design passed to you.

server/services/action.js
```diff
@@ -204,6 +208,26 @@ const actions = {
                         case actionTypes.ADD_CASE_NOTE:
                             await caseworkService.post(`/case/${caseId}/note`, { type: 'MANUAL', text: form.data['case-note'] }, headers);
                             break;
+                        case actionTypes.EXTEND_FOI_DEADLINE: {
```
SeanReedHomeOffice, line 211: It would have been nice to keep this more generic, with the type of extension passed as a parameter. Not sure if that is easy.

agraham19: I have addressed this.

server/services/action.js
```diff
@@ -204,6 +208,25 @@ const actions = {
...
+                                const response =
+                                    await caseworkService.post(`/case/${caseId}/stage/${stageId}/extension`,
+                                        { type: form.data.extensionType }, headers);
+
+                                const clientResponse = {
+                                    'summary': `Case: ${response.data.reference} extended`,
+                                    'link': `${response.data.reference}`
+                                };
+
+                                await caseworkService.post(`/case/${caseId}/note`,
```
SeanReedHomeOffice, line 222: Could the case note be added during the call to /case/${caseId}/stage/${stageId}/extension ?

SeanReedHomeOffice (changes requested):

server/services/action.js
agraham19: I've updated the applyExtension endpoint in casework to take a reason and create a note.

SeanReedHomeOffice (approved):
LGTM

PR #660, opened 2021-06-09
HOCS-2570 add new users for FOI approval teams

abarker66:
Added new users to keycloak to allow the progression of FOI workflow. 
Caseworker 2 and Caseworker 3 are now both members of FOI BICSPI G6G7 Approval team 1 and 
FOI BICSPI SCS Approval Team.


SeanReedHomeOffice (approved):
LGTM

PR #661, opened 2021-06-10
HOCS-3121 cutting max length of date entrys to proper sizes

gardnersam:
Maximum number of digits enterable set to 2 for day, 2 for month and 4 for year


LiamMacP:
Although this does fix the date component from passing through incorrect dates to validation. Is it also worth looking at the `isValidDate` validator too to improve that?

zuzak:
Not sure about this. If a user types "2020 01 01" into a fieldset, we shouldn't silently turn that into "20 01 0001".

Ideally we would figure out what they actually meant (e.g. accept "1 FEB 2020", chomp whitespace), and if we can't do that throw a validation message. In general you shouldn't restrict character lengths in fields: instead let the user type it in, and then tell them they're wrong with an error message (makes things like pasting easier, and it's confusing to a user who would expect to type anything in)

https://design-system.service.gov.uk/patterns/dates/
https://design-system.service.gov.uk/components/date-input/

----

I would also consider reformatting your Git commit messages to make it clear _why_ you're changing what you're changing: https://chris.beams.io/posts/git-commit/

gardnersam:
@LiamMacP I think the validator could be improved as you say, will look into it. The regex in the pattern attribute doesn't actually do anything as it doesn't apply to the "number" type. Could swap to "text" type and do it that way. 

@zuzak The Jira specifically asks for the number of enterable digits to be limited (as I understood it anyway). I think what you've described is basically the previous functionality, minus the ability to write Feb rather than 2, so the user could type 12345 in the month box and would receive a validation message saying please enter a number between 0 - 31.

gardnersam:
Complete change from what was previously peer reviewed, removed the html 5 validation from the dates and replaced it with checks and warnings matching the home office design guidance.
invalidmonth

server/middleware/validation.js
```diff
@@ -16,7 +16,10 @@ const validationErrors = {
     isValidWithinDate: label => `${label} must be within the last ${VALID_DAYS_RANGE} days`,
     validCaseReference: () => 'Case reference is not valid',
     contributionsFulfilled: () => 'Case contributions have to be completed or cancelled',
-    oneOf: () => 'Options are not valid'
+    oneOf: () => 'Options are not valid',
+    isValidMonth: () => 'Invalid month entered',
+    isValidYear: () => 'Invalid year entered',
+    isValidDay: () => 'Invalid day entered'
```
zuzak, line 22: You [shouldn't](https://design-system.service.gov.uk/components/error-message/#be-clear-and-concise) use "invalid" here because it doesn't add anything to the message. A good error message helps the user understand what went wong and how to fix it.

There's guidance on the "right" error messages for date validation -- see https://design-system.service.gov.uk/components/date-input/#if-the-date-entered-cannot-be-correct.

server/middleware/validation.js
```diff
@@ -16,7 +16,10 @@ const validationErrors = {
     isValidWithinDate: label => `${label} must be within the last ${VALID_DAYS_RANGE} days`,
     validCaseReference: () => 'Case reference is not valid',
     contributionsFulfilled: () => 'Case contributions have to be completed or cancelled',
-    oneOf: () => 'Options are not valid'
+    oneOf: () => 'Options are not valid',
```
zuzak, line 19: You shouldn't use "not valid" here. If this is an option group and the problem is that they didn't pick an option, use "Select a [whatever it is]". See: https://design-system.service.gov.uk/components/radios/#error-messages

server/middleware/validation.js
```diff
@@ -29,6 +32,24 @@ const validators = {
...
+        }
+        return null;
+    },
+    isValidMonth({ label, value, message }) {
+        if (getMonth(value) < 1 || getMonth(value) > 12) {
+            return message || validationErrors.isValidMonth(label);
+        }
+        return null;
+    },
+    isValidYear({ label, value, message }) {
```
zuzak, line 47: Not sure about this. 1700 _is_ a year, after all. It's just not a year we'd expect to see in correspondence metadata. This might break some edge-case date of birth fields: 100 years is still within a human lifespan.

Use a more precise name for this function. This isn't an `isValidYear()` -- it's an "is now±100y"
Consider using more precise message too: "Year must be after 1911".

server/middleware/validation.js
```diff
@@ -155,6 +176,18 @@ const validators = {
     }
 };
 
+function getDay(date){
+    return date.split('-')[2];
+}
```
zuzak, line 181: Consider:
```suggestion
const getDay(date) => date.split('-')[2];
```

You're inconsistently using
```js
function foo(bar){
// and
function foo(bar) {
```

Use whichever's already used in this file.

server/middleware/validation.js
```diff
@@ -29,6 +32,24 @@ const validators = {
...
+        return null;
+    },
+    isValidMonth({ label, value, message }) {
+        if (getMonth(value) < 1 || getMonth(value) > 12) {
+            return message || validationErrors.isValidMonth(label);
+        }
+        return null;
+    },
+    isValidYear({ label, value, message }) {
+        if (getYear(value) > (new Date().getFullYear() + 100) || getYear(value) < (new Date().getFullYear() - 100)) {
```
zuzak, line 48: Consider splitting this line over several lines. It's too long.

zuzak (commented):
Don't forget to tidy up your Git history.

server/middleware/validation.js
gardnersam: Updated and followed the "If the date entered cannot be correct" example as closely as possible. Ideally I think we would have a common message for all of these and use highlighting for more information as specified but due to the way the validator code is shared system wide I think changing it to enable this would have far reaching implications and not be worth it at this point

server/middleware/validation.js
gardnersam: Agreed and updated

server/middleware/validation.js
gardnersam: Good suggestion, updated.

server/middleware/validation.js
gardnersam: Updated

server/middleware/validation.js
```diff
@@ -16,7 +17,10 @@ const validationErrors = {
     isValidWithinDate: label => `${label} must be within the last ${VALID_DAYS_RANGE} days`,
     validCaseReference: () => 'Case reference is not valid',
     contributionsFulfilled: () => 'Case contributions have to be completed or cancelled',
-    oneOf: () => 'Options are not valid'
+    oneOf: () => 'Options are not valid',
+    isValidMonth: label =>  `${label} must contain a valid month`,
+    isYearWithinRange: label =>   `${label} must contain a valid year`,
+    isValidDay: label => `${label} must contain a valid day`,
```
zuzak, line 23: GOV.UK Design System just goes for "X must be a real date" and doesn't be more specific (as long as only the invalid day/month/year field gets highlighted in red). 

"Real" is probably understood better than "valid" by normal users.

server/services/forms/schemas/add-standard-line.js
```diff
@@ -14,6 +14,9 @@ module.exports = options => Form()
         Component('date', 'expiry_date')
             .withValidator('required', 'Expiration date is required')
             .withValidator('isValidDate', 'Expiration date must be a valid date')
+            .withValidator('isValidDay', 'Date must contain a valid day')
+            .withValidator('isValidMonth', 'Date must contain a valid month')
+            .withValidator('isYearWithinRange', 'Date must be before ' + (new Date().getFullYear() + 120))
```
zuzak, line 19: Magic number?

gardnersam: Good point, removed from here and several other places

server/middleware/validation.js
gardnersam: Due to the shared use of the validation code along with the factories to create forms by everything in the front end editing it to match the guidance exactly appears too high risk/effort to be valuable so deviating slightly from the defined wording to get a similar effect remained the only option. Agreed that "Real" will likely carry more meaning for people so reworded there.

server/services/forms/schemas/contribution-fulfillment.js
```diff
@@ -1,6 +1,7 @@
 const Form = require('../form-builder');
 const { Component, Choice, ConditionChoice } = require('../component-builder');
 const { getSomuItem } = require('../../../middleware/somu');
+const YEAR_RANGE = 120;
```
zuzak, line 4: Might be a bit neater to do this?
```suggestion
const MIN_ALLOWABLE_YEAR = (new Date().getFullYear() - YEAR_RANGE);
```

gardnersam: Agreed that makes it more readable, change made

LiamMacP:
You're referring to `YEAR_RANGE` in 5 different places with a value that is highly unlikely to change (due to it's meaning). Can this better be a environment variable? or pulled into a general config variable that can be shared? We have a [date helper class](https://github.com/UKHomeOffice/hocs-frontend/blob/4fba485774aac7058d8a381b4859dbdb7355a146/server/libs/dateHelpers.js) that might be a better place for it?

gardnersam:
Fair point about the reuse of YEAR_RANGE, moved it to dateHelper.js

server/middleware/validation.js
```diff
@@ -1,6 +1,6 @@
 const { FormSubmissionError, ValidationError } = require('../models/error');
 const { DOCUMENT_WHITELIST, DOCUMENT_BULK_LIMIT, VALID_DAYS_RANGE } = require('../config').forContext('server');
-const YEAR_RANGE = 120;
+const { YEAR_RANGE } = require('../libs/dateHelpers');
 const MIN_ALLOWABLE_YEAR = (new Date().getFullYear() - YEAR_RANGE);
```
LiamMacP, line 4: Work also moving these to a the dateHelpers? Feels like we are reusing these in multiple places too.

gardnersam: Yeah might aswell, I've put the min and max years into the helper and just expose them now

server/libs/dateHelpers.js
```diff
@@ -22,4 +22,10 @@ const parseDate = (rawDate) => {
...
 const formatDate = (date) => date ? parseDate(date) : null;
 
-module.exports = { addDays, getUtcDateString, formatDate };
+const YEAR_RANGE = 120;
+
+const MIN_ALLOWABLE_YEAR = (new Date().getFullYear() - YEAR_RANGE);
+
+const MAX_ALLOWABLE_YEAR = (new Date().getFullYear() + YEAR_RANGE);
+
+module.exports = { addDays, getUtcDateString, formatDate, MIN_ALLOWABLE_YEAR, MAX_ALLOWABLE_YEAR };
```
zuzak, line 31: Always end a file in Git with a newline. That way `diff` works better: https://unix.stackexchange.com/a/18789

server/middleware/validation.js
```diff
@@ -16,7 +17,10 @@ const validationErrors = {
     isValidWithinDate: label => `${label} must be within the last ${VALID_DAYS_RANGE} days`,
     validCaseReference: () => 'Case reference is not valid',
     contributionsFulfilled: () => 'Case contributions have to be completed or cancelled',
-    oneOf: () => 'Options are not valid'
+    oneOf: () => 'Options are not valid',
+    isValidMonth: label =>  `${label} must contain a real month`,
+    isYearWithinRange: label =>   `${label} must contain a valid year`,
```
zuzak, line 22: Don't use "valid".

server/middleware/validation.js
```diff
@@ -29,6 +33,30 @@ const validators = {
...
+            if (getMonth(value) < 1 || getMonth(value) > 12) {
+                return message || validationErrors.isValidMonth(label);
+            }
+        }
+        return null;
+    },
+    isYearWithinRange({ label, value, message }) {
+        if (value) {
+            if (getYear(value) > MAX_ALLOWABLE_YEAR || getYear(value) < MIN_ALLOWABLE_YEAR) {
+                return message || validationErrors.isYearWithinRange(label);
```
zuzak, line 55: Can you make it return `${label} must be after 1901` and `${label} must be before 2100` as appropriate?

server/libs/dateHelpers.js
gardnersam: Good point have now done here and will do everywhere else.

server/middleware/validation.js
gardnersam: Updated to "Select at least one option", this is currently only used for when the user selects a complaints type and the message is overridden with "Select at least one complaint type option".

server/middleware/validation.js
gardnersam: Yep change made

server/services/forms/schemas/add-standard-line.js
```diff
@@ -14,6 +15,9 @@ module.exports = options => Form()
         Component('date', 'expiry_date')
             .withValidator('required', 'Expiration date is required')
             .withValidator('isValidDate', 'Expiration date must be a valid date')
+            .withValidator('isValidDay', 'Date must contain a real day')
+            .withValidator('isValidMonth', 'Date must contain a real month')
+            .withValidator('isYearWithinRange', 'Date must be before ' + (new Date().getFullYear() + YEAR_RANGE))
```
zuzak, line 20: Will the user get this error if they enter 1782?

server/services/forms/schemas/bulk-document-add.js
```diff
@@ -1,13 +1,17 @@
...
 
 module.exports = () => Form()
     .withTitle('Create Bulk Cases')
     .withField(
         Component('date', 'DateReceived')
             .withValidator('required', 'Date received is required')
             .withValidator('isValidDate', 'Date received must be a valid date')
+            .withValidator('isValidDay', 'Date received must contain a real day')
+            .withValidator('isValidMonth', 'Date received must contain a real month')
+            .withValidator('isYearWithinRange', 'Date received must be after ' + MIN_ALLOWABLE_YEAR)
```
zuzak, line 14: Consider using template literals here (backtick strings)

server/services/forms/schemas/add-standard-line.js
gardnersam: No the validators are applied in order so they will get the next error "Expiration date must not be in the past" and only get the year validation error if they put something too far in the future

server/services/forms/schemas/bulk-document-add.js
gardnersam: Updated here and in single document add

src/shared/common/forms/date.jsx
```diff
@@ -72,10 +67,8 @@ class DateInput extends Component {
                                 name={monthKey}
                                 type="number"
                                 pattern="[0-9]*"
-                                min="1"
```
zuzak, line 75: I wonder if we should leave these in -- do they add semantic value?
(To disable the HTML5 validation we'll need to pass `<form novalidate>`, but I'm not sure we're ready to do that yet?)

server/middleware/validation.js
```diff
@@ -155,6 +187,10 @@ const validators = {
     }
 };
 
+const getDay = (date) => date.split('-')[2];
```
zuzak, line 190: Are there any scenarios wherein there aren't enough dashes in the date string?

src/shared/common/forms/__tests__/__snapshots__/date.spec.js.snap
```diff
@@ -656,4 +656,4 @@ exports[`Form date component should render with value when passed 1`] = `
     </div>
   </fieldset>
 </div>
-`;
+`;
```
zuzak, line 659: Always leave the last `\n` on final lines. You've accidentally removed it here -- remove this file from the commit.

server/middleware/validation.js
```diff
@@ -29,6 +34,33 @@ const validators = {
         }
         return null;
     },
+    isValidDay({ label, value, message }) {
+        if (value) {
+            if (getDay(value) > new Date(getYear(value), getMonth(value), 0).getDate() || getDay(value) < 1) {
```
zuzak, line 39: This will still accept `0001` as a date -- which is fine, it should -- but it should ideally transform silently to `01`. Might be another ticket though?

zuzak (commented):
I don't really like how we're passing `MIN_ALLOWABLE_YEAR` all over the place, but I think it's the best we can do?

I wonder if `isYearWithinRange` should be renamed to something clearer, but I'm not sure what it'd be

server/middleware/validation.js
gardnersam: Going to leave this for now, can raise a ticket if we want to later

src/shared/common/forms/__tests__/__snapshots__/date.spec.js.snap
gardnersam: Good spot added back in

server/middleware/validation.js
gardnersam: In the current usage no, the date string comes through as undefined if not all are present so it doesn't reach this stage, however in the interest of reusable code I have updated these methods to check length

zuzak (approved):
I don't really like how we're passing `MIN_ALLOWABLE_YEAR` everywhere, but it'll do the job.

PR #665, opened 2021-06-14
HOCS-3127 - add sort strategy for primaryCorrespondent

robertdeniszczyc2:


gardnersam (approved):
Verified locally, both columns sort now

PR #666, opened 2021-06-14
HOCS-3128: change radio control ids

LiamMacP:
At present there are instances whereby the choices have values that
contain characters that are not allowed in CSS selectors. This change
improves the id that the radio uses so that there is a reduced chance
of these characters being part of the selectors. We now use the
name of the character and the key index to identify each of the
options. Due to this change various tests need to be updated to use
this new id.

### DIFF
--- a/src/shared/common/forms/radio-group.jsx
+++ b/src/shared/common/forms/radio-group.jsx
@@ -103,6 +103,10 @@ class Radio extends Component {
         return choicesToUse;
     }
 
+    getIdName(nameOfField, key) {
+        return `${nameOfField}-${key}`;
+    }
+
     render() {
         const {
             className,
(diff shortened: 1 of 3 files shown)


src/shared/common/forms/radio-group.jsx
```diff
@@ -103,6 +103,10 @@ class Radio extends Component {
         return choicesToUse;
     }
 
+    getIdName(nameOfField, key) {
+        return `${nameOfField}-${key}`;
+    }
```
zuzak, line 108: dunno if this is a good idea, but:
```suggestion
    getIdName(nameOfField, key) => `${nameOfField}-${key}`;
```

gardnersam (approved):
Fix worked locally, page refreshed properly with no blank screen as seen previously

src/shared/common/forms/radio-group.jsx
LiamMacP: This is following the existing pattern - doesn't make any real difference.

PR #667, opened 2021-06-14
HOCS-3103 Adding contributions overdue indication to complaints triag…

gardnersam:
…e workstack

### DIFF
--- a/server/lists/adapters/workstacks.js
+++ b/server/lists/adapters/workstacks.js
@@ -173,6 +173,23 @@ const bindDisplayElements = fromStaticList => async (stage) => {
         }
     }
 
+    if (stage.data && stage.data.CaseContributions){
+        JSON.parse(stage.data.CaseContributions).forEach(contribution => {
+            if(contribution.data.contributionDueDate && new Date(contribution.data.contributionDueDate) <= new Date() && !contribution.data.contributionStatus){
+               stage.contributions = 'Overdue';
+            }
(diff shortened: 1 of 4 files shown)


server/lists/adapters/workstacks.js
```diff
@@ -173,6 +173,25 @@ const bindDisplayElements = fromStaticList => async (stage) => {
...
+                else if(contribution.data.contributionDueDate && !contribution.data.contributionStatus && stage.contributions != 'Overdue'){
+                    stage.contributions = 'Due';
+                }
+                else if(contribution.data.contributionStatus == 'contributionCancelled' && stage.contributions != 'Due' && stage.contributions != 'Overdue'){
+                    stage.contributions = 'Cancelled';
+                }
+                else if(contribution.data.contributionStatus == 'contributionReceived' && stage.contributions != 'Due' && stage.contributions != 'Cancelled'  && stage.contributions != 'Overdue'){
+                    stage.contributions = 'Complete';
+                }
+            }
```
robertdeniszczyc2, line 191: I have some concerns about the maintainability of this approach, I think there may be cleaner ways to approach the problem potentially using filter(), and a way to avoid having to do all the not checks. I think there is a similar thing to this happening just above in the same file around line 158.

robertdeniszczyc2 (changes requested):

robertdeniszczyc2 (approved):
Looks good to me

PR #670, opened 2021-06-16
HOCS-3193 Sticky Cases

SeanReedHomeOffice:
If at the end of a stage the user is a member of the team that handles the case at the next stage, then the case might want to be be allocated to them automatically and displayed on screen, instead of just going to the workstack.

When the workflow service - updateStageForward() - detects that the end of a stage is reached - in getStage() - it checks if the current user is on the team for the next stage. If they are the user is allocated to that case, and data relating to the new stage is returned to the front end.
The front end code detects if the stageId has changed, and uses that instead when it redirects.

The feature can be turned on by setting a process variable in the Camunda diagram. 

sticky_cases001

### DIFF
--- a/server/services/action.js
+++ b/server/services/action.js
@@ -66,9 +66,9 @@ async function handleActionSuccess(response, workflow, form) {
     }
 }
 
-function handleWorkflowSuccess(response, { caseId, stageId }) {
+function handleWorkflowSuccess(response, { caseId }) {
     if (response.data && response.data.form) {
-        return { callbackUrl: `/case/${caseId}/stage/${stageId}` };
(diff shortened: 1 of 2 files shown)


server/services/action.js
```diff
@@ -233,6 +237,7 @@ const actions = {
         }
     },
     WORKFLOW: async ({ caseId, stageId, form, user }) => {
+        console.log('server/services/action.js');
```
agraham19, line 240: Console.log left over

gary-newbolt (approved):
lgtm

LiamMacP (approved):
Looks good to me, needs to be rebased off main 👍🏽

PR #671, opened 2021-06-21
HOCS-2994 Update ReactJS to 17.0.2

robertdeniszczyc2:


package-lock.json
```diff
@@ -6520,7 +6596,7 @@
       "dependencies": {
         "combined-stream": {
           "version": "1.0.6",
-          "resolved": "https://registry.npmjs.org/combined-stream/-/combined-stream-1.0.6.tgz",
+          "resolved": "http://registry.npmjs.org/combined-stream/-/combined-stream-1.0.6.tgz",
```
zuzak, line 6599: Your commit's changed lots of registry links from HTTPS to HTTP, which isn't supposed to happen.
(It's safe to `sed` this manually to https://, or use `git commit -p` to only commit the changes you want)

robertdeniszczyc2: Good spot had missed that in the diff when I committed, will fix

zuzak (approved):
That lockfile diff is a mess, but seems a lot of it is just indentation and order changes

PR #672, opened 2021-06-22
HOCS-3079 Alter application banner when running in a NOTPROD environment

fewtrees-ho:
https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-3079

### DIFF
--- a/src/shared/common/components/__tests__/phase-banner.spec.js
+++ b/src/shared/common/components/__tests__/phase-banner.spec.js
@@ -17,4 +17,9 @@ describe('Phase Banner component', () => {
             render(<PhaseBanner feedback="http://some.domain" />)
         ).toMatchSnapshot();
     });
+    it('should display the correct test notice when NOTPROD', () => {
+        expect(
+            render(<PhaseBanner environment="NOTPROD" />)
+        ).toMatchSnapshot();
+    });
 });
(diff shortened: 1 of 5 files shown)


src/shared/common/components/phase-banner.jsx
```diff
@@ -3,14 +3,21 @@ import PropTypes from 'prop-types';
 
 export default class PhaseBannerComponent extends Component {
     render() {
-        const { feedback, phase } = this.props;
+        const { feedback, phase, environment } = this.props;
+
+        var envDisplay;
```
robertdeniszczyc2, line 8:
```suggestion
        let envDisplay;
```

src/shared/common/components/__tests__/phase-banner.spec.js
```diff
@@ -17,4 +17,9 @@ describe('Phase Banner component', () => {
             render(<PhaseBanner feedback="http://some.domain" />)
         ).toMatchSnapshot();
     });
+    it('should display the correct test notice when NOTPROD', () => {
+        expect(
+            render(<PhaseBanner environment="NOTPROD" />)
+        ).toMatchSnapshot();
+    });
```
robertdeniszczyc2, line 24: Also add a test for displaying the correct message when prod

robertdeniszczyc2 (changes requested):
Added a couple minor comments

src/shared/common/components/phase-banner.jsx
fewtrees-ho: changed to let in the new code

src/shared/common/components/__tests__/phase-banner.spec.js
fewtrees-ho: Added explicit test for this - there was a default test but explicit is better

robertdeniszczyc2 (approved):
Looks good to me

PR #675, opened 2021-06-24
HOCS-2423 : DCU draft teams

MarkWrightBjss:
Redirect drafting teams request to new endpoint that gets teams by stage type.


server/lists/index.js
```diff
@@ -293,7 +293,7 @@ module.exports = {
         },
         DRAFT_TEAMS: {
             client: 'INFO',
-            endpoint: '/teams/drafters',
```
zuzak, line 296: Is there a `/teams/drafters` endpoint we should tear out?

zuzak: ah, you already have. Don't mind me

philbarton (approved):
Looks good to me

server/lists/index.js
MarkWrightBjss: You're absolutely right. See here:
https://github.com/UKHomeOffice/hocs-info-service/pull/306/files

SeanReedHomeOffice (approved):
LGTM

SeanReedHomeOffice (approved):
Maybe should be called DCU_DRAFT_TEAMS to avoid confusion. But not a biggie

PR #676, opened 2021-06-24
HOCS-3103 Using new data from somu table rather than case data to sho…

gardnersam:
…w contribution status

### DIFF
--- a/server/lists/adapters/__tests__/workstacks.spec.js
+++ b/server/lists/adapters/__tests__/workstacks.spec.js
@@ -998,13 +998,13 @@ describe('decorateContributionsWithStatus', () => {
     it('returns the decorated contributions with statuses', () => {
         const currentDate = new Date('2021-06-18 12:30');
 
-        var contributionOne = '{"contributionRequestDate":"2021-06-18","contributionDueDate":"2021-06-02","contributionStatus":"contributionReceived"}';
-        var contributionTwo = '{"contributionDueDate":"2021-06-20","contributionRequestDate":"2021-06-18"}';
-        var contributionThree = '{"contributionDueDate":"2021-06-02","contributionRequestDate":"2021-06-18"}';
-        var contributionFour = '{"contributionDueDate":"2021-06-02","contributionRequestDate":"2021-06-18","contributionStatus":"contributionCancelled"}';
-        var array = '{"caseContributions":[]}';
+        const contributionOne = '{"contributionRequestDate":"2021-06-18","contributionDueDate":"2021-06-02","contributionStatus":"contributionReceived"}';
+        const contributionTwo = '{"contributionDueDate":"2021-06-20","contributionRequestDate":"2021-06-18"}';
+        const contributionThree = '{"contributionDueDate":"2021-06-02","contributionRequestDate":"2021-06-18"}';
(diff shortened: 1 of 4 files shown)


philbarton (commented):
This looks ok to me but might be worth asking Liam to have a quick look.

src/shared/common/components/workstack.jsx
```diff
@@ -387,13 +387,13 @@ class WorkstackAllocate extends Component {
             case ColumnRenderer.TRUNCATE_TEXT:
                 return <td key={row.uuid + column.dataValueKey} className='govuk-table__cell govuk-table__cell--truncated' title={value}>{value}</td>;
             case ColumnRenderer.CONTRIBUTIONS_WARNING:
-                if (row.data.CaseContributions) {
-                    const dueContribution = JSON.parse(row.data.CaseContributions)
-                        .filter(contribution => contribution.data && !contribution.data.contributionStatus)
-                        .map(contribution => contribution.data.contributionDueDate)
+                if (row.somu && row.somu.caseContributions) {
+                    const dueContribution = row.somu.caseContributions
+                        .filter(contribution => !JSON.parse(contribution).contributionStatus)
```
LiamMacP, line 392: Could you do a map previous to this to do the `JSON.parse` and then filter afterwards? Means you don't repeat the parse?

server/lists/adapters/workstacks.js
```diff
@@ -125,23 +125,24 @@ const highestPriorityContributionStatus = (decoratedContributions) => {
     let highestPriority = 0;
 
     decoratedContributions.map(decoratedContribution => {
```
LiamMacP, line 127: This should probably be a `forEach`, map creates a new array and is for the transformation of data.

LiamMacP: Could also use `reduce` if you are feeling fancy...

server/lists/adapters/__tests__/workstacks.spec.js
```diff
@@ -996,54 +996,61 @@ describe('Workflow Workstack Adapter', () => {
...
-           '{"data":{"contributionDueDate":"2021-06-02","contributionRequestDate":"2021-06-18","contributionStatus":"contributionCancelled"}}]';
         const currentDate = new Date('2021-06-18 12:30');
 
-        const result = decorateContributionsWithStatus(JSON.parse(contributions), currentDate);
-
-        expect(result[0].data['contributionStatus']).toEqual('contributionReceived');
-        expect(result[1].data['contributionStatus']).toEqual('contributionDue');
-        expect(result[2].data['contributionStatus']).toEqual('contributionOverdue');
-        expect(result[3].data['contributionStatus']).toEqual('contributionCancelled');
+        var contributionOne = '{"contributionRequestDate":"2021-06-18","contributionDueDate":"2021-06-02","contributionStatus":"contributionReceived"}';
```
zuzak, line 1001: Use `let` or `const`, not `var`.

gardnersam: Updated

server/lists/adapters/workstacks.js
gardnersam: Agreed, gone for forEach, apologies for reduced fanciness levels

src/shared/common/components/workstack.jsx
gardnersam: Yep think it increases readability

LiamMacP (approved):

PR #677, opened 2021-06-30
HOCS-3315: utilise show stack trace environment var

LiamMacP:
At present to show stack traces within the error we rely on the
`isProduction` value in the config page. As turning this off in
non-local environments causes unintended consequences. To
mitigate this, a new `SHOW_STACKTRACE_ON_ERROR_PAGE` env
variable is being passed through. This is then used to signal whether
to pass through the stacktrace to the error model.

### DIFF
--- a/server/config.js
+++ b/server/config.js
@@ -1,6 +1,7 @@
 const isProduction = process.env.NODE_ENV === 'production';
 const workflowAuth = (process.env.WORKFLOW_BASIC_AUTH || 'UNSET:UNSET').split(':');
 const isNotProd = process.env.IS_NOTPROD === '1';
+const showStackTraceInErrorPage = process.env.SHOW_STACKTRACE_ON_ERROR_PAGE === '1';
 
 const config = {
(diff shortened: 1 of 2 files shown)


server/config.js
```diff
@@ -1,6 +1,7 @@
 const isProduction = process.env.NODE_ENV === 'production';
 const workflowAuth = (process.env.WORKFLOW_BASIC_AUTH || 'UNSET:UNSET').split(':');
 const isNotProd = process.env.IS_NOTPROD === '1';
+const showStackTraceInErrorPage = process.env.SHOW_STACKTRACE_ON_ERROR_PAGE === '1';
```
zuzak, line 4: Default this to the value of `isNotProd`.

LiamMacP: It defaults to false if it's not there anyway? `isNotProd` is true for demo environments that contradicts what is in the AC of the ticket.

zuzak (approved):

PR #679, opened 2021-07-07
HOCS-3236 Double encoding slashes to be compatible with hocs-casework…

gardnersam:
… update

### DIFF
--- a/server/routes/dashboard.js
+++ b/server/routes/dashboard.js
@@ -10,6 +10,7 @@ const { bindDisplayElements } = require('../lists/adapters/workstacks');
 const getLogger = require('../libs/logger');
 const { caseworkService } = require('../clients');
 const User = require('../models/user');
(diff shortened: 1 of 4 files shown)


server/routes/dashboard.js
```diff
@@ -29,7 +29,7 @@ router.post(['/search/reference', '/api/search/reference'],
 
             logger.info('SEARCH_REFERENCE', { reference: caseRef });
 
-            const reference = encodeURIComponent(caseRef);
+            const reference = encodeURIComponent(caseRef).split('%2F').join('%252F');
```
philbarton, line 32: Could we wrap this in a function and test?

philbarton (changes requested):
Looks good, but change as discussed.

philbarton (approved):
lgtm

PR #684, opened 2021-07-12
Feature/hocs 3172 mpam see mp in workstack

gardnersam:


src/shared/common/components/workstack.jsx
```diff
@@ -317,6 +319,15 @@ class WorkstackAllocate extends Component {
                     }
                     <Link to={`/case/${row.caseUUID}/stage/${row.uuid}`} className='govuk-link govuk-!-margin-right-3'>{value.caseReference}</Link>
                 </td>;
+            case ColumnRenderer.MP_WITH_OWNER:
+                return <td key={row.uuid + column.dataValueKey} className='govuk-table__cell'>
+                    {
+                        <span className='govuk-!-font-weight-bold'>{value.mp}<br/></span>
```
zuzak, line 325: Don't  use <br> for presentational purposes. Consider using the grid that comes with a design system, or setting the MP name to be a block element.

Using an override class here's a bit of a red flag -- is there a better design we could use? For example, a [heading with caption](https://design-system.service.gov.uk/styles/typography/#headings-with-captions).  Should we use `<strong>` here instead of span?

gardnersam: I'll sort out the <br> usage. I'd hesitate to swap to a different styling as this matches the rendering for this one which is used a fair bit around DCU:

> case ColumnRenderer.CORRESPONDENT_WITH_CASE_LINK:  
                  return <td key={row.uuid + column.dataValueKey} className='govuk-table__cell'>
                    {
                        value.primaryCorrespondentFullName &&
                        <span className='govuk-!-font-weight-bold'>{value.primaryCorrespondentFullName}<br/></span>
                    }
                    <Link to={`/case/${row.caseUUID}/stage/${row.uuid}`} className='govuk-link govuk-!-margin-right-3'>{value.caseReference}</Link>
                </td>;

I'll try the strong tag and see if it looks out of place.

gardnersam: Updated the tags to get rid of the <br> and overriding, result looks identical so I am happy with it.

philbarton (approved):
lgtm

PR #686, opened 2021-07-14
HOCS-3316 Only flushing old static lists if retrieval of new ones is …

gardnersam:
…successful

This fixes a bug where occasionally team names would not appear on the dashboard.

Team names are stored in static lists which are refreshed every 15 minutes. This bug occurred when the refresh failed due to the info or casework services not being available as the old behaviour was to delete the static lists, request new ones and repopulate when they arrived, so failure to retrieve new ones left us with nothing. The new behaviour waits to see if new lists are successfully retrieved before deleting and replacing the old data. In the event of failure the preexisting static lists are maintained and the user won't know anything is amiss.

### DIFF
--- a/server/services/list/__tests__/service.spec.js
+++ b/server/services/list/__tests__/service.spec.js
@@ -502,16 +502,53 @@ describe('List Service', () => {
         });
     });
 
-    describe('when the flush method is called', () => {
-        it('should call the flush on the instance', async () => {
+    describe('when the flush method is called but retreival of new data fails', () => {
+        it('should call not call flush or store on the respository', async () => {
             const listKey = '__listKey__';
+
             const mockRepositoryFlush = jest.fn();
+            const mockRepositoryStore = jest.fn();
+            const mockRepositoryFetch = jest.fn();
+            const mockClientInstance = jest.fn();
+            mockClientInstance.get = mockGet;
+            mockRepositoryFetch.mockReturnValue(mockClientInstance);
+            const mockGet = jest.fn();
+            mockGet.mockImplementation(() => {
+                throw new Error('400');
+            });
 
-            await listService.initialise({}, {}, { listRepository: { flush: mockRepositoryFlush } });
+            await listService.initialise({}, {}, { listRepository: { flush: mockRepositoryFlush, fetch: mockRepositoryFetch, store: mockRepositoryStore } });
 
-            listService.flush(listKey);
+            await listService.flush(listKey);
 
             expect(getLogger).toHaveBeenCalled();
+            expect(mockRepositoryFlush).not.toHaveBeenCalled();
+            expect(mockRepositoryStore).not.toHaveBeenCalled();
+        });
+    });
+
+    describe('when the flush method is called and retreival of new data is successful', () => {
+        it('should call call flush and store on the repository', async () => {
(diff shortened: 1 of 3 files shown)


LiamMacP:
How does this work if user 1 is loading a dashboard and finds no team and while it's flushing the cache user 2 loads the dashboard. Will we have infinite loop of flushing and reloading? We had this issue before which is why we went down the cronjob route.

I don't know if a better route would be to treat this more like a map, we replace if its there, and then flush the values that haven't been found in the static list response?

gardnersam:
@LiamMacP The flush/reload is still only triggered by the cronjob. Users loading the dashboard has no effect on the process. The change here is when the cronjob triggers the refresh the flush is only done if new data has been retrieved and the update is then handled in the same way as before.

LiamMacP:
> @LiamMacP The flush/reload is still only triggered by the cronjob. Users loading the dashboard has no effect on the process. The change here is when the cronjob triggers the refresh the flush is only done if new data has been retrieved and the update is then handled in the same way as before.

Ah my bad, yeah makes sense! Ignore me, nothing to see here...

server/services/list/service.js
```diff
@@ -173,24 +173,23 @@ const getInstance = (requestId, user) => {
...
 const cacheStaticList = async (listId) => {
     const { endpoint, client, adapter } = listRepository.fetch(listId);
     const clientInstance = clientRepository.fetch(client);
     const logger = getLogger();
-    let response;
     try {
-        response = await clientInstance.get(endpoint);
+        const response = await clientInstance.get(endpoint);
+        listCache.flush(listId);
+        const listData = await applyAdapter(response.data, adapter, { logger });
```
LiamMacP, line 187: Worth moving this up a line, I doubt the applying of the adapter will take long, but would remove the ability to flush and then the adapter to fail...

server/services/list/__tests__/service.spec.js
```diff
@@ -502,16 +502,53 @@ describe('List Service', () => {
...
+            const mockRepositoryFetch = jest.fn();
+            const mockGet = jest.fn();
+            mockGet.mockReturnValue(JSON.parse('{"client": "newclient"}'));
+
+            const mockClientInstance = jest.fn();
+            mockClientInstance.get = mockGet;
+            mockRepositoryFetch.mockReturnValue(mockClientInstance);
 
-            await listService.initialise({}, {}, { listRepository: { flush: mockRepositoryFlush } });
+            await listService.initialise({}, {}, { listRepository: { flush: mockRepositoryFlush, fetch: mockRepositoryFetch, store: mockRepositoryStore } });
```
zuzak, line 544: Split this over several lines.
```suggestion
            await listService.initialise({}, {}, { listRepository: {
                fetch: mockRepositoryFetch,
                flush: mockRepositoryFlush,
                store: mockRepositoryStore
            } });
```

gardnersam: Done for this and the other similar line

server/services/list/service.js
gardnersam: Good shout, I've made the change

fewtrees-ho (approved):
Approved as per teams call

PR #687, opened 2021-07-15
HOCS-3393 FOI Consider and Draft Stage - Unable to 'Add a Contribution'

agraham19:
* move FOI contributions block to the correct json block


gary-newbolt (approved):
lgtm

PR #688, opened 2021-07-15
HOCS-3397 Issues caused by team deactivation through Management UI

agraham19:
Deactivated team workstacks are appearing on the dashboard with blank labels. These labels are populated by the frontend from a list of teams from the info-service. Currently only the list of active teams is used, causing the issue.

* Use the full team list in the list service, including deactivated teams

### DIFF
--- a/server/lists/adapters/workstacks.js
+++ b/server/lists/adapters/workstacks.js
@@ -147,7 +147,7 @@ const decorateContributionsWithStatus = (contributions, currentDate) => {
 };
 
 const bindDisplayElements = fromStaticList => async (stage) => {
-    stage.assignedTeamDisplay = await fromStaticList('S_TEAMS', stage.teamUUID);
+    stage.assignedTeamDisplay = await fromStaticList('S_ALL_TEAMS', stage.teamUUID);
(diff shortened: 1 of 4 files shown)


LiamMacP (approved):

server/lists/adapters/workstacks.js
```diff
@@ -147,7 +147,7 @@ const decorateContributionsWithStatus = (contributions, currentDate) => {
 };
 
 const bindDisplayElements = fromStaticList => async (stage) => {
-    stage.assignedTeamDisplay = await fromStaticList('S_TEAMS', stage.teamUUID, true);
+    stage.assignedTeamDisplay = await fromStaticList('S_ALL_TEAMS', stage.teamUUID, true);
```
philbarton, line 150: Need to make sure that true is removed from these calls.

philbarton (changes requested):

server/lists/adapters/workstacks.js
agraham19: I've addressed this.

philbarton (approved):
lgtm

PR #689, opened 2021-07-16
HOCS-3008 : Re-apply stopForceFlush

philbarton:
This change re-applies a fix that was reverted inadvertently during a merge from comp to main.


robertdeniszczyc2 (approved):
Change matches original change performed under HOCS-3008

PR #694, opened 2021-07-21
HOCS-3424 - Accept response compression.

philbarton:
Accept compressed response.

### DIFF
--- a/server/libs/request.js
+++ b/server/libs/request.js
@@ -4,6 +4,8 @@ const fs = require('fs');
 
 const isProduction = process.env.NODE_ENV === 'production';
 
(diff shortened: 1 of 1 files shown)


fewtrees-ho (approved):

server/libs/request.js
```diff
@@ -4,6 +4,8 @@ const fs = require('fs');
 
 const isProduction = process.env.NODE_ENV === 'production';
 
+axios.defaults.headers.common['Accept-encoding'] = 'gzip';
```
zuzak, line 7: it should be case insensitive but just in case:
```suggestion
axios.defaults.headers.common['Accept-Encoding'] = 'gzip';
```

PR #699, opened 2021-07-22
HOCS-3311 Workstack Tags

robertdeniszczyc2:
Adds support for displaying tags in workstack column data cells, to indicate particular statuses about a case clearly on a workstack view.

### DIFF
--- a/server/lists/adapters/workstacks.js
+++ b/server/lists/adapters/workstacks.js
@@ -42,6 +42,36 @@ const defaultCaseSort = (a, b) => {
     }
     return sortResult;
 };
+
+/**
+ * Used in tagSort compare function.
+ * a.tag and b.tag are arrays, we compare lengths to determine if a case has tags.
+ * If a case has tags, it has a higher precedence in a workstack so is moved higher in the rendered list in the UI.
+ * If both cases being compared have tags, or do not have tags, they have the same precedence so we don't alter the
+ * ordering.
+ */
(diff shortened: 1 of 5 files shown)


src/shared/common/components/workstack-tag.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function Tags(props) {
+    let tags = [];
+
+    if (props.row.tag && props.row.tag.length > 0) {
+        tags = props.row.tag.map(tag => {
+            return <span key={tag} className="govuk-tag govuk-!-margin-right-1">{tag}</span>;
```
zuzak, line 9: consider `<abbr>`

robertdeniszczyc2: Currently we only have one tag which is an abbreviation I guess but not sure if we guarantee/enforce all future tags should be abbreviations?

server/lists/adapters/workstacks.js
```diff
@@ -42,6 +42,29 @@ const defaultCaseSort = (a, b) => {
     }
     return sortResult;
 };
+
+const byTag = (a, b) => {
```
gardnersam, line 46: I had to look at this for a while to understand what it's doing but am not sure why it is doing it. Seems like a very specific version of sorting based on length. Maybe needs more explanation through variable names or comments?

PR #701, opened 2021-07-26
HOCS-3408 - Ability to send an initial response

agraham19:
* Add initial response document list to list service


gary-newbolt (approved):
lgtm

PR #704, opened 2021-07-30
HOCS-3013: change dashboard endpoint call

LiamMacP:
To support the new result set provided by the `/dashboard`
endpoint in casework, this change parses the result that will 
be returned. As we no longer need computation (as this is 
done in the backend service), we can simply just populate 
each of the cards.

This change also changes the list that is used when hitting 
the dashboard to `/dashboard` of casework.


philbarton (approved):
Looks good and runs locally.

PR #705, opened 2021-07-30
HOCS-3477 - FOI Approvals

agraham19:
* Add contributions style forms for Approvals


server/middleware/validation.js
```diff
@@ -17,13 +17,38 @@ const validationErrors = {
...
+            });
+
+            if (result.length !== 0) {
+                valid = false;
+            }
+        } else {
+            valid = false;
+        }
+    } catch (error) {
+        valid = false;
```
gary-newbolt, line 46: Any reason why we aren't logging out the error here? Seems to me the likely culprit would be malformed JSON response which is an error we would want to know about.

gary-newbolt (commented):
Just that one thing on here. Thanks

server/middleware/validation.js
agraham19: I've now added a logging statement here.

gary-newbolt: great thanks

gary-newbolt (approved):

PR #712, opened 2021-08-02
Feature/hocs 3233 unworkable cases

philbarton:
WCS Unworkable feature.

If a case is marked "Unworkable" it should be hidden, apart from in search.
This change adds a filter to workstack data, and adds a search by "Unworkable cases only".

### DIFF
--- a/server/lists/adapters/workstacks.js
+++ b/server/lists/adapters/workstacks.js
@@ -20,6 +20,22 @@ const byPriority = (a, b) => {
     }
 };
 
+const byWorkable = (stage) => {
+    if (stage.data) {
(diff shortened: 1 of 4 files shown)


server/lists/adapters/workstacks.js
```diff
@@ -20,6 +20,22 @@ const byPriority = (a, b) => {
     }
 };
 
+const byWorkable = (stage) => {
+    if (stage.data) {
```
gardnersam, line 24: Could probably combine these first two if statements, but not a deal breaker

gardnersam (approved):
Looks good

PR #715, opened 2021-08-03
HOCS-3480 Create component for Correspondence Details in the Review D…

agraham19:
…ata Table of Case Creation stage

* Handle the primary correspondent in the review-field


gary-newbolt (approved):
lgtm

PR #716, opened 2021-08-03
HOCS-3503 Remove FOI Type field from Case creation and Check your ans…

agraham19:
…wers screen as it is no more required for MVP

* Remove FOI type field


gary-newbolt (approved):
lgtm

PR #720, opened 2021-08-06
HOCS-3474 Sorting on Requester/Reference column is broken in all work…

agraham19:
…stacks

This issue is called by old cases which do not have correspondents. This
change adds a check to stop these throwing errors.

* Check primaryCorrespondent is defined before trying to access fullname


gary-newbolt (approved):
lgtm

PR #722, opened 2021-08-09
HOCS 3417 Allocate a case

LaurieBennett:
Updates the allocation stage, removing the Request Type Selection screen, removing the Allocation Case Note entry field and adding in a confirmation screen displaying the team name and case number.

### DIFF
--- a/server/services/action.js
+++ b/server/services/action.js
@@ -94,7 +94,6 @@ const actions = {
                     case actionTypes.CREATE_CASE: {
                         const { data: documentTags } = await getDocumentTags(context);
                         response = await createCase('/case', { caseType: context, form }, documentTags[0], headers);
-                        console.log('responseData', response.data);
                         clientResponse = { summary: 'Case Created ', link: `${response.data.reference}` };
                         return handleActionSuccess(clientResponse, workflow, form);
                     }
--- a/server/services/action.js
+++ b/server/services/action.js
@@ -94,6 +94,7 @@ const actions = {
                     case actionTypes.CREATE_CASE: {
                         const { data: documentTags } = await getDocumentTags(context);
                         response = await createCase('/case', { caseType: context, form }, documentTags[0], headers);
+                        console.log('responseData', response.data);
                         clientResponse = { summary: 'Case Created ', link: `${response.data.reference}` };
                         return handleActionSuccess(clientResponse, workflow, form);
                     }
--- a/src/shared/common/forms/__tests__/confirmation-with-team-name-and-case-ref.spec.js
+++ b/src/shared/common/forms/__tests__/confirmation-with-team-name-and-case-ref.spec.js
@@ -1,0 +1,38 @@
+import React from 'react';
+import ConfirmationWithTeamNameAndCaseRef from '../confirmation-with-team-name-and-case-ref.jsx';
+import { ApplicationProvider } from '../../../contexts/application.jsx';
+
+describe('Panel component', () => {
+    const config = {
+        csrf: '__token__'
+    };
+
+    const choices = [
+        {
+            key: '4ef92480-922f-11eb-a8b3-0242ac130003',
+            value: 'FOI BICSPI Acceptance Team'
+        }
+    ];
+
+    const data = {
+        AcceptanceTeam: '4ef92480-922f-11eb-a8b3-0242ac130003'
+    };
+
+    xit('should render with default props', () => {
+        expect(
+            render(<ApplicationProvider>
+                <ConfirmationWithTeamNameAndCaseRef />
+            </ApplicationProvider>)
+        ).toMatchSnapshot();
+    });
+
+
+    it('should render with title and team name when passed', () => {
+        expect(
+            render(<ApplicationProvider config={config} data={data}>
+                <ConfirmationWithTeamNameAndCaseRef data={data} choices={choices} label="completed" caseRef={'test-case-ref'} />
+            </ApplicationProvider>)
+        ).toMatchSnapshot();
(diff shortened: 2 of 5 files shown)


server/services/action.js
```diff
@@ -94,6 +94,7 @@ const actions = {
                     case actionTypes.CREATE_CASE: {
                         const { data: documentTags } = await getDocumentTags(context);
                         response = await createCase('/case', { caseType: context, form }, documentTags[0], headers);
+                        console.log('responseData', response.data);
```
gary-newbolt, line 97: Can you clean this one up please?

src/shared/common/forms/form-repository.jsx
```diff
@@ -159,6 +160,12 @@ export function formComponentFactory(field, options) {
             return renderFormComponent(Inset, { key, data, config });
         case 'confirmation-with-case-ref':
             return renderFormComponent(ConfirmationWithCaseRef, { key, data, config, caseRef });
+        case 'confirmation-with-team-name-and-case-ref':
+            var teamName;
```
gary-newbolt, line 164: any reason for the `var` over `const`?

src/shared/common/forms/form-repository.jsx
```diff
@@ -159,6 +160,12 @@ export function formComponentFactory(field, options) {
...
+            if (data.AcceptanceTeam){
```
gary-newbolt, line 165: Can this logic be moved into the component? It will keep the switch a little cleaner. I know there are already some instances where someone has added logic in the case block but really it's polluting the class a little and you seem to be passing all that's needed to the component anyway.

gary-newbolt (changes requested):
You really need a test class on your new `ConfirmationWithCaseRef` component, similar to the other tests on other components... checking the rendering under different prop data conditions tested against snapshots.

src/shared/common/forms/form-repository.jsx
LaurieBennett: Ah, the reason I kept this outside was due to potential to extend this with other team types, the team name may not always be passed through an AcceptanceTeam variable, so I figured this was the best way to keep it usable in other functionalities. Can still change it if you want.

src/shared/common/forms/form-repository.jsx
LaurieBennett: The intent was to allow the team name to be set based on any number of potential variables, not just for AcceptanceTeam, I could potentially move the setter and the return value inside of the if and use a const if this would be preferred?

src/shared/common/forms/form-repository.jsx
gary-newbolt: Hi @LaurieBennett , I still think this logic should be moved into the component for the same reason state above. I also think that rather than passing the `AcceptanceTeam` variable into the component, extract the team from `data` once inside the component. That follows the pattern of the method. 

In the component, itself the `AcceptanceTeam` should be extracted. I accept your point that this component should be usable by teams other than `AcceptanceTeams` so a solution will need to be worked out. A potential solution is to filter the `data` object for the type of team required based on the `CurrentStage`, and use a key-value pair to say... if current stage is X then I'm looking for a team type of Y. Give me a shout if you want any help.

src/shared/common/forms/form-repository.jsx
gary-newbolt: I think this is redundant if given the request below.

src/shared/common/forms/confirmation-with-team-name-and-case-ref.jsx
```diff
@@ -0,0 +1,47 @@
+import React, { Component, Fragment } from 'react';
+import PropTypes from 'prop-types';
+
+export default class ConfirmationWithCaseRef extends Component {
```
gary-newbolt, line 4: Also spotted after the initial review... the name of this class needs to be changed and should read `ConfirmationWithTeamNameAndCaseRef`

src/shared/common/forms/confirmation-with-team-name-and-case-ref.jsx
```diff
@@ -0,0 +1,47 @@
...
+            <div>
+                <div className="govuk-panel govuk-panel--confirmation">
+                    <div className="govuk-panel__body">Case {label}: {teamName}
+                    </div>
+                </div>
+                <h2 className="govuk-heading-m">
+                    Case {caseRef} {label} {teamName}
+                </h2>
+                <div>
+                    {nextActions &&
```
gary-newbolt, line 25: It does look like you're not using this at all as well, again is it there for a reason?

LaurieBennett: It was part of the initial spec, to have a number of next actions, we couldn't work out a satisfactory way to do this easily within this case, I'll remove it for now

src/shared/common/forms/__tests__/confirmation-with-team-name-and-case-ref.spec.js
```diff
@@ -0,0 +1,38 @@
...
+            key: '4ef92480-922f-11eb-a8b3-0242ac130003',
+            value: 'FOI BICSPI Acceptance Team'
+        }
+    ];
+
+    const data = {
+        AcceptanceTeam: '4ef92480-922f-11eb-a8b3-0242ac130003'
+    };
+
+    xit('should render with default props', () => {
```
gary-newbolt, line 21: disabled test

gary-newbolt (approved):
that's great... thanks @LaurieBennett

PR #723, opened 2021-08-09
HOCS-3525 - Case Timeline not updated when correspondent updated

agraham19:
* Add CORRESPONDENT_UPDATED type to case note renderer


gary-newbolt (approved):
lgtm

PR #729, opened 2021-08-18
Feature/hocs 3547 overdue rework

philbarton:
This change removes the need for somu data being passsed to the frontend for contribution calculations.
Contribution dates are now calculated in casework, and added to the stage data.
For COMP the contributions stage is also calculated in casework. (I left MPAM stage type display as is, because its not part of this ticket and hard to unpick, plus no somu data used)
Also removed calls to add and edit contribution data in the case data blob.


gardnersam (approved):
Tested locally, works well

LiamMacP:
Code looks good, couple of things just to clean up the codebase now we are using the data from the `somu_data` table.

The following lines can be removed:
https://github.com/UKHomeOffice/hocs-frontend/blob/e5decab8c614805815021cc5ea0d0d109c8b2b6f/server/services/action.js#L215
https://github.com/UKHomeOffice/hocs-frontend/blob/e5decab8c614805815021cc5ea0d0d109c8b2b6f/server/services/action.js#L221

Alongside this, as we now have the one source of truth in the above table, we don't need to save somu-list component values:
https://github.com/UKHomeOffice/hocs-frontend/blob/e5decab8c614805815021cc5ea0d0d109c8b2b6f/server/middleware/process.js#L86 would become something like `.filter(field => field.type !== 'display' && field.type !== 'somu-list')`. That could be cleaned up to be in a function. 

Doing these changes would also fix the following ticket: HOCS-2631

LiamMacP (changes requested):
Requested a couple of changes that will clean up unused code, improve the data transfer and simplify the workflow.

philbarton:
> Code looks good, couple of things just to clean up the codebase now we are using the data from the `somu_data` table.
> 
> The following lines can be removed:
> 
> https://github.com/UKHomeOffice/hocs-frontend/blob/e5decab8c614805815021cc5ea0d0d109c8b2b6f/server/services/action.js#L215
> 
> https://github.com/UKHomeOffice/hocs-frontend/blob/e5decab8c614805815021cc5ea0d0d109c8b2b6f/server/services/action.js#L221
> 
> Alongside this, as we now have the one source of truth in the above table, we don't need to save somu-list component values:
> 
> https://github.com/UKHomeOffice/hocs-frontend/blob/e5decab8c614805815021cc5ea0d0d109c8b2b6f/server/middleware/process.js#L86
> would become something like `.filter(field => field.type !== 'display' && field.type !== 'somu-list')`. That could be cleaned up to be in a function.
> 
> Doing these changes would also fix the following ticket: HOCS-2631

I've made the change process.js. The puts that you commented on were already removed .

LiamMacP (approved):
Approved looks good! 🙂

PR #731, opened 2021-08-19
HOCS-3008 Prevent force flushing of cache (#639)

gary-newbolt:
Cache flushing was being triggered in various places and under
particular circumstances resulted in many flush requests being triggered unintentionally
in a very short space of time.

This change removes the force flush, so the cache clear will stick to the
cron schedule.


gary-newbolt:
Cherry picked the original code from main branch rather than duplicating.

abarker66 (approved):
LGTM.

PR #733, opened 2021-08-19
HOCS-3645: add correspondent type display name

LiamMacP:
As a result of now returning the correspondent display name as part of
the request. If it is present we should use it, otherwise just the type
should be shown. This change also moves the whole title 'generator' to
within a function.


gardnersam (approved):
Looks good

PR #737, opened 2021-08-25
HOCS-3670: user workstack simplification

LiamMacP:
Modify the USER_WORKSTACK list to point to the new endpoint in casework.
This endpoint passes through the user uuid.

To utilise the new endpoint result from casework, this change adapts the
 `userAdapter` to remove the unecessary filters. Alongside this an
 improvement to the byCaseReference comparator to check if the value is
 present.


gardnersam (approved):
Looks good, nice to have some of the other methods tidied up aswell

PR #743, opened 2021-08-31
HOCS-2089 Adding label to text box

gardnersam:
Orphaned form label issue fixed by using the inputId attribute for the Select component, this results in the label defined on line 73 being properly associated with the input.


philbarton (approved):
LGTM

PR #747, opened 2021-09-02
HOCS-3699 - Browser back throws error 'Something has gone wrong' on c…

agraham19:
…licking browser back button

* Replace get descendants by current team with get descendants by team uuid

### DIFF
--- a/server/lists/index.js
+++ b/server/lists/index.js
@@ -239,7 +239,7 @@ module.exports = {
         },
         FOI_DRAFT_TEAMS_SELECTION: {
             client: 'INFO',
-            endpoint: '/teams/descendants/stage/${stageId}/case/${caseId}',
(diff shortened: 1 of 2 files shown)


abarker66:
Andy I have built DECS using your branches and tried the back button in the listed places and it works the first time but if you attempt a second back button press I am seeing 500 http codes in the frontend logging and Camunda exceptions in the workflow service?! looks a bit broken.

agraham19:
> Andy I have built DECS using your branches and tried the back button in the listed places and it works the first time but if you attempt a second back button press I am seeing 500 http codes in the frontend logging and Camunda exceptions in the workflow service?! looks a bit broken.

I've rebased this now. You should be able to follow the whole case flow through.

abarker66 (approved):
All looks good, re tested and works fine.

PR #751, opened 2021-09-07
HOCS-3733 - Appeals are not recorded in timeline

agraham19:
* Send create note requests to casework when appeals created/updated


gary-newbolt (approved):
lgtm

PR #753, opened 2021-09-09
HOCS-3732: fix for post creation topic fix

gary-newbolt:


abarker66 (approved):
LGTM - the parent map flattening maybe useful for something else I noticed, unrelated.

PR #756, opened 2021-09-10
HOCS-3730 add Organisation field

abarker66:
Added an Organisation field underneath the requester's name field.


LaurieBennett (approved):
Looks Good, Pass

PR #757, opened 2021-09-14
HOCS-3766 Adding option to autofil date field with current date

gardnersam:


ARobinson26:
Change looks good. Is there any way to test this? I'm not sure how you could do it as the date in the snapshot would be need to be different everyday you ran the test. Is it possible to mock `Date()`?

src/shared/common/forms/date.jsx
```diff
@@ -33,12 +33,20 @@ class DateInput extends Component {
...
-            value
+            value,
+            autopopulate
         } = this.props;
         const yearKey = this.datePart('year');
         const monthKey = this.datePart('month');
         const dayKey = this.datePart('day');
-        const parts = this.parseValue(value);
+        let parts;
+        if(!value && autopopulate){
```
LiamMacP, line 43: `autopopulate` of value 'false' would be truthy here?

gardnersam: You're right it was evaluating to truthy regardless of value, I've swapped to a boolean rather than String in data now and it works properly

gardnersam:
@ARobinson26 I've added tests for autofill and no autofill. The Date test already had functionality to override the current date luckily.

src/shared/common/forms/date.jsx
LiamMacP: Not to overly conflate this, but I do see future issues with people passing Strings instead of booleans through in data.

Worth adding a check like this? https://github.com/UKHomeOffice/hocs-frontend/blob/2bbe63530d7d73ab95c3df78fdc9991626de7b1e/src/shared/common/forms/radio-group.jsx#L110

Happy to be veto'd if you think that this is sufficient.

gardnersam: I agree with what you're saying, I think the counter argument would be that this defaults to not being used, so really people aren't going to do like autopopulate = false because it is false anyway. And if they put a String it should be obvious pretty quickly that they've done it wrong?

LiamMacP: My feeling would be whether it is obvious? all other props apart from the one above is all string based. Although it might be obvious that somethings wrong, is it obvious what the fix is? I see that people are just going to look over `"false"` and `false`...

LiamMacP (approved):
Tests have been added and comments resolved, looks good!

PR #786, opened 2021-10-27
hocs 3925 refactor actions

gary-newbolt:


server/middleware/case.js
```diff
@@ -91,6 +93,58 @@ function caseCorrespondentsApiResponseMiddleware(req, res) {
...
 }
 
+async function caseActionDataMiddleware(req, res, next) {
+    try {
+        const preppedData = {};
+        const actionData = await req.listService.fetch('CASE_ACTIONS', req.params);
+        let uniqueActionTypeIds;
+        let uniqueActionTypes;
+
+        if (actionData && actionData.caseTypeActionData) {
```
agraham19, line 103: These first two ifs both have the same condition `(actionData && actionData.caseTypeActionData)`.

If both of these variables must be defined for the successful execution of the function, could we check for their presence at the start, then throw and log a meaningful error if they're missing? We could then get rid of the ifs.

server/routes/api/case.js
```diff
@@ -74,6 +87,8 @@ router.put('/:caseId/note/:noteId',
 
 router.get('/:caseId/summary', caseSummaryMiddleware, caseSummaryApiResponseMiddleware);
 
+router.get('/:caseId/actions', caseActionDataMiddleware, caseActionApiResponseMiddleware);
```
agraham19, line 90: This single caseActionDataMiddleware function is used for both gets and posts and seems to be quite complex. It looks like it might have multiple responsibilities. Would we get simpler methods if we split it up?

gary-newbolt: With both the `GET` and `POST` requests the role of this middleware is the same thing, i.e. to get the current info. So to separate it in this instance I think would only lead to more duplicate code.

server/middleware/case.js
gary-newbolt: I'll alter to combine the IF statements... it needs to be able to successfully return even if there is no caseActionData, I see no need to throw. I'll add a log message at warn level to advise no info found.

server/services/forms/index.js
```diff
@@ -547,10 +543,10 @@ module.exports = {
         }
     },
     getFormForCase: async (options) => {
```
agraham19, line 545: Would it be possible to have a `getFormForCase` for each specialised use? I think it might be easier to read if we had methods such as `getActionFormForCase` .etc

agraham19 (approved):
Approved with a few comments/questions

server/services/forms/index.js
gary-newbolt: After discussion decided to leave as is.

PR #797, opened 2021-11-09
HOCS-4058: fix display of appeals in summary

gary-newbolt:


abarker66 (approved):
Well it appears to work so all good. Tests pass.

PR #805, opened 2021-11-19
HOCS-3824 Update clearance fulfilment page for MPAM

gardnersam:


server/middleware/validation.js
```diff
@@ -280,6 +281,15 @@ const validators = {
             }
         }
         return message || validationErrors.oneOf();
+    },
+    isValidWithinNextThirtyDays({ label, value, message }) {
```
robertdeniszczyc2, line 285: Wonder if there's a way we can make this a generic isValidWithinDays validator where we can pass in the number of days?

I'm thinking if on the field we can set a prop on the field and access that in the validation middleware to pass through into the validator.

Also requires tests adding in `hocs-frontend/server/middleware/__tests__/validation.spec.js`

robertdeniszczyc2 (changes requested):

server/middleware/validation.js
gardnersam: @robertdeniszczyc2 Good idea, I've swapped it to a generic version, see what you think. Also added tests.

server/middleware/validation.js
```diff
@@ -17,6 +17,7 @@ const validationErrors = {
     isBeforeToday: label => `${label} must be a date in the past`,
     isAfterToday: label => `${label} must be a date in the future`,
     isValidWithinDate: label => `${label} must be within the last ${VALID_DAYS_RANGE} days`,
+    isValidWithinNextThirtyDays: label => `${label} must be within the next 30 days`,
```
robertdeniszczyc2, line 20: Don't think this is required now?

```suggestion
```

server/middleware/validation.js
```diff
@@ -280,6 +282,14 @@ const validators = {
             }
         }
         return message || validationErrors.oneOf();
+    },
+    isValidWithinGivenDays({ label, value, message, props }) {
+        let limitDate = new Date();
+        limitDate.setDate(limitDate.getDate() + parseInt(props));
```
robertdeniszczyc2, line 288: Would props _only_ contain a value for this, or could there be potential it would have more fields? In which case you'd access the property on the object instead

server/middleware/__tests__/validation.spec.js
```diff
@@ -450,6 +450,18 @@ describe('Validators', () => {
             expect(validators.oneOf({ submittedFormData: { 'delay': 'delay' }, options: options, message: 'test error message' })).toEqual(null);
         });
     });
+
+    describe('Valid within given days validator', () => {
+        it('should reject if more than 30 days after current date', () => {
```
robertdeniszczyc2, line 455: Would make the descriptions on these more generic, to avoid an inference it only works for 30 days.

i.e. should reject if more days than the given days allowed

Also apply same for the test on line 460

robertdeniszczyc2 (changes requested):

server/middleware/validation.js
gardnersam: Nope, removed

server/middleware/__tests__/validation.spec.js
gardnersam: Good shout, updated

server/middleware/validation.js
gardnersam: The props can be anything, it'll be passed in on the validator object so if people need this sort of thing in the future they can use props in whatever way works for them

robertdeniszczyc2 (approved):

PR #806, opened 2021-11-19
HOCS-3977 - Case Details accordion missing some sections for FOI case

agraham19:
* Fix issue which causes case view accordions to appear in the wrong error


gary-newbolt (approved):
lgtm

PR #807, opened 2021-11-23
Fix/hocs 4060 deadline validation foi

gary-newbolt:


abarker66 (changes requested):
During testing I noticed an edge case where a new case with a full 20 days deadline a user will not see any options in the days drop down list which is a required field. There is no way to progress from this. Not likely to happen in practise but it can happen.

abarker66 (approved):
If they read the instructions it could save a few phone calls, all good.

PR #808, opened 2021-11-23
HOCS-4201 'ICO Review' appeal type appearing as 'Ico Review' in summa…

agraham19:
…ry tab

* Remove title case operation


gary-newbolt (approved):
lgtm

PR #815, opened 2021-11-25
HOCS-3821: Update MPAM add clearance request page

JosephRogerson:


server/middleware/validation.js
```diff
@@ -289,6 +290,14 @@ const validators = {
             return message || validationErrors.isValidWithinGivenDays(label, props);
         }
         return null;
+    },
+    isValidWithinPastGivenDays({ label, value, message, props }) {
+        let limitDate = new Date();
```
LiamMacP, line 295:
```suggestion
        const limitDate = new Date();
```

server/middleware/validation.js
```diff
@@ -26,6 +26,7 @@ const validationErrors = {
     isAfterMinYear: label => `${label} must be after ${MIN_ALLOWABLE_YEAR}`,
     isValidDay: label => `${label} must contain a real day`,
     isValidWithinGivenDays: (label, props) => `${label} must not be more than ${props} days in the future`,
+    isValidWithinPastGivenDays: (label, props) => `${label} must not be more than ${props} days in the past`,
```
LiamMacP, line 29:
```suggestion
    isValidWithinPastGivenDays: (label, props) => `${label} must be within the last ${props} days.`,
```
I'm not overly fussed, but this feels less wordy.

server/middleware/__tests__/validation.spec.js
```diff
@@ -462,6 +462,17 @@ describe('Validators', () => {
         });
     });
 
+    describe('Valid within past given days validator', () => {
+        it('should reject if more than given days before current date', () => {
+            let inputDate = new Date();
```
LiamMacP, line 467:
```suggestion
            const inputDate = new Date();
```

server/middleware/validation.js
```diff
@@ -289,6 +290,14 @@ const validators = {
             return message || validationErrors.isValidWithinGivenDays(label, props);
         }
         return null;
+    },
+    isValidWithinPastGivenDays({ label, value, message, props }) {
```
LiamMacP, line 294: I know this is the current use case but would having the props part be `{ days: 30 }` be more explicit. Refering as a prop doesn't give us the context.

This would then become the following:

```suggestion
    isValidWithinPastGivenDays({ label, value, message, { days }) {
```

server/middleware/validation.js
LiamMacP: Might be worth looking at the following: https://design-system.service.gov.uk/components/error-message/

server/middleware/validation.js
JosephRogerson: Makes sense, would `isValidWithinGivenDays` also need changing to `days` so that they both work in the same way?

LiamMacP: I'd say so, would need to make sure that the fields are updating that use the `isValidWithinGivenDays` as well.

fewtrees-ho (approved):

PR #820, opened 2021-12-02
HOCS-4103 - PIT Extensions

agraham19:
* Disallow PIT Extensions after the case deadline


abarker66 (commented):
Nearly all AC is met : There is a line 'Its done when a correspondence type of 'PIT Extension' is able to be uploaded when applying a PIT extension.' is this done? How do we upload docs for PIT extensions?
Every time a case note is generated for Extensions there is a spurious extra case note (System Event) created as well - looks buggy.

agraham19:
> Nearly all AC is met : There is a line 'Its done when a correspondence type of 'PIT Extension' is able to be uploaded when applying a PIT extension.' is this done? How do we upload docs for PIT extensions? Every time a case note is generated for Extensions there is a spurious extra case note (System Event) created as well - looks buggy.

I've added an upload document option to the create extension itself now - previously you had to do it through the Manage Document tab.

The notes seem to be working okay now. I think it's just because the branch didn't have the fix in.

abarker66 (approved):
Looks good, confirmed docs can now be uploaded for extension.

PR #830, opened 2021-12-08
Feature/hocs 4123 to triage stage

gary-newbolt:


server/lists/index.js
```diff
@@ -676,7 +676,145 @@ module.exports = {
             endpoint: '/entity/list/BF_ENQUIRY_REASON',
             type: listService.types.STATIC,
             adapter: entityListItemsAdapter
-        }
+        },
```
agraham19, line 679: I think it would be better to have a single parameterised entityList list, but I'll approve as this is the pattern we've been following.

agraham19 (approved):

PR #835, opened 2021-12-15
HOCS-4198 Extra blank case note when adding/updating appeals and 'Sys…

agraham19:
…tem Event' log in timeline when an extension is applied

* Stop saving case notes for actions
* Rely on the audit timeline for action notes


abarker66 (commented):
I think there could be a cut & paste error in case-notes.js line 178, update not create?

agraham19:
> I think there could be a cut & paste error in case-notes.js line 178, update not create?

I've updated this now;

abarker66 (approved):
Typo fixed, LGTM. System tested - all passed.

PR #836, opened 2021-12-20
HOCS-4345 add organisation field to show in people and summary tabs

abarker66:
Added organisation field to show in people and case summary tabs.


agraham19 (approved):
I've tested this and it works

PR #846, opened 2022-01-17
HOCS-3391: Remove undefined from COMP accordion

JosephRogerson:
There is currently a check for undefined businessArea to stop an
undefined business area being displayed, this change adds another check
for an undefined business unit.

### DIFF
--- a/server/lists/adapters/case-view-all-stages.js
+++ b/server/lists/adapters/case-view-all-stages.js
@@ -40,7 +40,12 @@ const getRequestStatus = (dueDate, { status, decision }) => {
 
 async function composeBusinessLabel(choices, { businessArea, businessUnit }, fromStaticList) {
     const businessUnitLabel = await loadValue(businessUnit, choices, fromStaticList);
-    return businessArea ? `${businessArea} - ${businessUnitLabel}` : businessUnitLabel;
+    if (businessArea && businessUnitLabel) {
+        return `${businessArea} - ${businessUnitLabel}`;
+    }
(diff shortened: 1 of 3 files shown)


server/lists/adapters/case-view-all-stages.js
```diff
@@ -40,7 +40,12 @@ const getRequestStatus = (dueDate, { status, decision }) => {
 
 async function composeBusinessLabel(choices, { businessArea, businessUnit }, fromStaticList) {
     const businessUnitLabel = await loadValue(businessUnit, choices, fromStaticList);
-    return businessArea ? `${businessArea} - ${businessUnitLabel}` : businessUnitLabel;
+    if (businessArea && businessUnitLabel) {
+        return `${businessArea} - ${businessUnitLabel}`;
+    }
+    else {
+        return businessArea ? businessArea : businessUnitLabel;
+    }
```
robertdeniszczyc2, line 48: Could add a test for this logic. 

Give me a shout if you need a hand :)

Cheers,
Rob

robertdeniszczyc2 (commented):
Just one comment re potential for adding a test

Cheers,
Rob

server/lists/adapters/case-view-all-stages.js
JosephRogerson: I've added a test, so both execution paths should be covered now.

Thanks,
Joseph

robertdeniszczyc2 (approved):

PR #847, opened 2022-01-24
HOCS-4395: BF: Add actions tab

JosephRogerson:


src/shared/common/components/summary/actions/action-summary-interests.jsx
```diff
@@ -0,0 +1,45 @@
...
+        return false;
+    }
+
+    return (
+        <>
+            <h2 className='govuk-heading-m'>Interest</h2>
+            { interests.map((item, index) => (
+                <>
+                    <table className='govuk-table margin-left--small'>
+                        <caption className='govuk-table__caption margin-bottom--small'>Interest { index + 1 }</caption>
```
robertdeniszczyc2, line 32: Is this index + 1 bit of this used for anything or purely display purposes? If it's used as a reference for business reasons such as an identifier we should use a different method of determining the identifier as this in frontend component is potentially unstable. Happy to talk through for further context

server/lists/adapters/actions/externalInterestsSummaryAdapter.js
```diff
@@ -0,0 +1,12 @@
+module.exports = async (actions) => {
+
+    const { recordInterest: interests } = actions.caseActionData;
+
+    return await Promise.all(interests.map( async interestEntry => {
```
robertdeniszczyc2, line 5: This breaches Sonarlint rule: "await" should not be used redundantly

An async function always wraps the return value in a Promise. Using return await is therefore redundant.

robertdeniszczyc2 (changes requested):
Added a few comments, give me a shout with any questions

Cheers,
Rob

src/shared/common/components/summary/actions/action-summary-interests.jsx
JosephRogerson: It's just for display purposes. It's not used as a reference for anything, we just went for this as it was an easy way to separate the interests in the summary tab

src/shared/common/components/summary/actions/action-summary-factory.jsx
```diff
@@ -10,10 +11,11 @@ function renderComponent(Component, items) {
 }
 
 const getActionSummary = (actionType, actions) => {
-
     switch (actionType) {
         case 'Appeals':
             return renderComponent(AppealSummary, actions.find(actions => actions.title === actionType));
+        case 'Recordinterest':
```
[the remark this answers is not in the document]
LiamMacP: Might be an idea asking @gary-newbolt about this - I'm not sure I know what the reason would be from the top of my head for this. I agree this would probably be nicer pascal cased.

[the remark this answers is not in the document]
gary-newbolt: That was probably missed on the original code review for the creation of the `RecordInterest` action... it's worth changing for sure, you would probably need to make a corresponding change in `hocs-casework` off of the top of my head, it's worth a check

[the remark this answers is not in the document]
LiamMacP: In the database it appears to be `recordInterest` and we then change to `Recordinterest`. I think its just a change to the way the action type is handled in the frontend. If you need a hand with this @JosephRogerson give me a shout!

src/shared/common/components/summary/actions/action-summary-interests.jsx
```diff
@@ -0,0 +1,45 @@
...
+    value: PropTypes.string.isRequired
+};
+
+const ActionSummaryInterests = (props) => {
+
+    const interests = [ ...props.props.items ];
+
+    if (interests.length < 1) {
+        return false;
+    }
```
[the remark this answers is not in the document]
JosephRogerson: I've changed this to pass in the arrays to both appeals and expected interests and added items to the propTypes. 

I'm happy to change it from false, but I don't think we want to display a message if there are no values. I would expect that if there are no interests then it would not appear in the summary tab.

src/shared/common/components/summary/actions/action-summary-factory.jsx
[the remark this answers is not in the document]
JosephRogerson: I've changed this now so that we compare against the value that is passed to the frontend from the database rather than the formatted value

src/shared/common/components/summary/actions/action-summary-appeals.jsx
```diff
@@ -12,12 +12,13 @@ const renderRow = ({ label, value }) => {
...
     label: PropTypes.string.isRequired,
-    value: PropTypes.string.isRequired
+    value: PropTypes.string.isRequired,
+    items: PropTypes.array.isRequired
 };
 
 const ActionSummaryAppeals = (props) => {
 
-    const appeals = [ ...props.props.items ];
+    const { items: appeals } = props;
```
LiamMacP, line 21: Nitpick: this could be put directly into the props passthrough

src/shared/helpers/case-type-sidebar-tabs-flags.js
```diff
@@ -13,6 +13,9 @@ const caseTypeSidebarTabsFlags = {
     FOI: {
         people: false,
         foi_actions: true
+    },
+    BF: {
+        foi_actions: true
```
LiamMacP, line 18: I know this is unrelated and a side effect of how actions was genericised. Could you raise a ticket to generalise this please...

src/shared/common/components/summary/actions/action-summary-interests.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+    return (
+        <>
+            <h2 className='govuk-heading-m'>Interest</h2>
+            { interests.map((item, index) => (
+                <>
+                    <table className='govuk-table margin-left--small'>
```
LiamMacP, line 32: I do wonder if pulling Interest out into its own component would be beneficial? @robertdeniszczyc2 what's your opinion?

src/shared/helpers/case-type-sidebar-tabs-flags.js
JosephRogerson: I've raised a ticket to change this to just `actions`

src/shared/common/components/summary/actions/action-summary-interests.jsx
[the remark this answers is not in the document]
robertdeniszczyc2: Makes sense - I think when I commented this I'd not considered fully it was the summary tab

src/shared/common/components/summary/actions/action-summary-interests.jsx
robertdeniszczyc2: I guess if it's just for display purposes does it add anything to the UX and therefore is it required? I'm just thinking it could be confusing for the user/users might assume it's an identifier and start using it as a reference elsewhere, for example "in interest 2 there is xyz...".

Might be worth running past interaction designer?

src/shared/common/components/summary/actions/action-summary-interests.jsx
robertdeniszczyc2: Yeah that could work, small single responsibility components are always preferable :)

JosephRogerson: I've lifted this and the corresponding appeals code into their own components

src/shared/common/components/summary/actions/action-summary-interests.jsx
JosephRogerson: I've removed the index + 1 as part of the lifting to a separate component, as I can see how it would be confusing and isn't consistent with appeals

robertdeniszczyc2 (dismissed):

robertdeniszczyc2 (approved):

PR #860, opened 2022-02-01
HOCS-4468 Additional Document Types for Treat Official

agraham19:
* Update Final Response documents list endpoint url


gary-newbolt (approved):
lgtm

PR #872, opened 2022-02-11
HOCS-4542 Support multiple case types in case note adapter

robertdeniszczyc2:
External interest case note adapter supports multiple case types


LiamMacP (approved):
Change looks good - feels weird though

robertdeniszczyc2:
@LiamMacP cheers, I agree - I also felt that when I made the change. I think the type adapter might be doing too much "stuff", maybe opportunity to refactor some of this behaviour

robertdeniszczyc2:
Talked through the change with @LiamMacP and agree refactoring this functionality can be picked up separately as out of scope of this immediate change.

PR #884, opened 2022-02-18
HOCS-4578 Type param when updating case data

robertdeniszczyc2:
Adds support for passing a type query parameter when updating case data.
Used when updating case data and a further upstream action needs to be
performed such as creating a case note.

Also adds support in timeline for EX_GRATIA_UPDATE case note type.


gary-newbolt (approved):
lgtm

PR #892, opened 2022-03-02
Nm sort test

nickmcmahon01:


nickmcmahon01:
Our largest team is over 8000 cases now, there are multiple teams with hundreds of cases now.
Small change to replace localCompare for content we control (and so don't need to worry about locale) for a more basic search.
but very small change and should reduce TTFB for larger teams by up to 600ms..
https://stackoverflow.com/questions/14677060/400x-sorting-speedup-by-switching-a-localecompareb-to-ab-1ab10

PR #896, opened 2022-03-07
Feature/hocs 4566 add reference field to case summary

abarker66:
Added a correspondent reference field to the case summary.


gary-newbolt (approved):
lgtm

PR #909, opened 2022-03-18
HOCS-3736: create new teams and unit dropdown for foi

gary-newbolt:


src/shared/common/forms/__tests__/confirmation-with-team-name-and-case-ref.spec.js
```diff
@@ -11,18 +11,17 @@ describe('Panel component', () => {
         {
             key: '4ef92480-922f-11eb-a8b3-0242ac130003',
             value: 'FOI BICSPI Acceptance Team'
+        },
+        {
+            key: 'abc92480-922f-11eb-a8b3-0242ac130003',
+            value: 'Another Team'
```
LaurieBennett, line 17: Was this intended to be left in here?

LaurieBennett (changes requested):

src/shared/common/forms/__tests__/confirmation-with-team-name-and-case-ref.spec.js
gary-newbolt: Hi, yes it was,  just so under the hood it verifies it's matching the key, value pairs correctly.

PR #911, opened 2022-03-18
HOCS-4632 Add GRO Drafting Teams list

robertdeniszczyc2:


LiamMacP:
Downscaling this from a Major to Minor - default behaviour if unit doesn't exist is info returns 0 teams.

LiamMacP (approved):

PR #913, opened 2022-03-21
HOCS-4774: pass through labels to document

LiamMacP:
[].concat returns a new label and does not modify the existing arrays
that are merged. This change returns the value of the concat.


zuzak (approved):
Another case for Typescript, I guess.

PR #915, opened 2022-03-22
HOCS-4398 BF2 add interest to action tab

abarker66:
added capability for BF2 case officer to record interest.


gary-newbolt (approved):
lgtm

PR #920, opened 2022-03-28
HOCS-4837 Add matrix to test against latest-LTS of NodeJS as well as prod version

zuzak:
We're currently deploying an explicit version of NodeJS, which will inevitably end up being older than we'd like. This PR  changes our pipeline to also test against the latest LTS release of NodeJS, so that when we update we're definitely sure it'll work.

Most of the time the two builds will be running against the same binary. I considered adding the latest-latest version of NodeJS (at time of writing, Node 18) but I think that added overhead to the build that we don't need to consider: we're not going to upgrade for a few years yet.

This has a "failure: ignore" step on the new pipeline for now, as the LTS version of NodeJS breaks our build. We'll remove it as part of the LTS upgrade work.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -11,26 +11,27 @@ trigger:
 
 steps:
   - name: install dependencies
-    image: node:14.15.4-alpine
+    image: &node-version
+      node:14.15.4-alpine
     commands:
       - npm --loglevel warn install --production=false --no-optional
 
   - name: build project
-    image: node:14.15.4-alpine
+    image: *node-version
     commands:
       - npm run build-prod
     depends_on:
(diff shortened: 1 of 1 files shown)


zuzak:
This PR will fail the label checker because "skip-release" is invalid; https://github.com/UKHomeOffice/hocs-frontend/pull/921 fixes that.

robertdeniszczyc2:
Does this run in parallel to the build?

zuzak:
Yes; you can check by looking at the "card view" at the top of the Drone build.

robertdeniszczyc2 (approved):

PR #921, opened 2022-03-28
Allow "skip-release" as a valid pull request label

zuzak:
I have a commit that doesn't contain any code changes, but at the moment
I'm forced to bump a version number for the app even though that doesn't
make sense.

This commit reconfigures the label check to also allow "skip-release", a
label that already exists on this repository.


LiamMacP:
This will need a change to the semver tagging action to ignore pull requests with this tag. Will sort it out tomorrow.

PR #937, opened 2022-04-14
HOCS-4954: bump UUID to latest

LiamMacP:
Bump UUID library to current latest version 8.3.2.


robertdeniszczyc2:
The diff in the package-lock.json looks unexpectedly large. Might be worth double checking it's what we expect from updating one dependancy?

robertdeniszczyc2 (approved):

PR #957, opened 2022-04-28
HOCS-4982 Retire Enzyme for RTL

robertdeniszczyc2:
Retire Enzyme in favour of React Testing Library.

The React 18 release changes the React API and entrypoint into the application. This API change is incompatible with Enzyme.

Enzyme has been dying for a while, but looks to be completely dead now (No commits since September) and the contributor who made the React 17 adapter to continue support post React 16 has said there will be no React 18 adapter: https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl

This removes Enzyme and brings in React Testing Library.

This PR includes:

- The update of React to React 18
- Removal of Enzyme, runner, config
- Conversion of tests to React Testing Library. Some tests have new matchers implemented, others updated snapshots.
- Tests which were found to be wrong have been fixed
- Tests which were testing implementation against the spirit of RTL core concepts (such as testing for the presence of CSS classes on a component) have been removed.


src/shared/common/forms/__tests__/__snapshots__/type-ahead.spec.js.snap
```diff
@@ -440,6 +440,7 @@ exports[`Form type ahead component (select) should render disabled when passed 1
                 id="type-ahead"
```
robertdeniszczyc2, line 440: I'm not sure what the Noop is introduced here - this is an Enzyme generated snapshot running in legacy mode via React 18.

LiamMacP (approved):
Will need a rebase and probably have some merge issues...

robertdeniszczyc2:
Looks like the rebase update has meant the commits are duplicated. Might be easier for me to just take the diff and apply onto a new branch; tried rebasing and it conflicts on every commit.

PR #959, opened 2022-05-03
HOCS-4409: Render tabs from DB

JosephRogerson:


src/shared/common/components/side-bar.jsx
```diff
@@ -59,19 +61,25 @@ class SideBar extends Component {
...
-                        {this.renderTabButton('Timeline', 'TIMELINE')}
-                        {caseTabs.people && this.renderTabButton('People', 'PEOPLE')}
-                        {caseTabs.case_actions && this.renderTabButton('Actions', 'CASE_ACTIONS')}
-                        {caseTabs.ex_gratia && this.renderTabButton('Ex-Gratia', 'EX_GRATIA')}
+                        {caseTabs.includes('documents') && this.renderTabButton('Documents', 'DOCUMENTS')}
+                        {caseTabs.includes('summary') && this.renderTabButton('Summary', 'SUMMARY')}
+                        {caseTabs.includes('timeline') && this.renderTabButton('Timeline', 'TIMELINE')}
+                        {caseTabs.includes('people') && this.renderTabButton('People', 'PEOPLE')}
+                        {caseTabs.includes('actions') && this.renderTabButton('Actions', 'FOI_ACTIONS')}
+                        {caseTabs.includes('ex_gratia') && this.renderTabButton('Ex-Gratia', 'EX_GRATIA')}
```
robertdeniszczyc2, line 82: By default on line 19 the active tab is set to documents if there is no active tab. Would be worth checking what happens if documents is not a tab allowed on the case type being viewed

JosephRogerson: Good spot, it renders the document tab regardless. I think it should probably default to the left-most tab instead

JosephRogerson: I've changed it so that it will pick the leftmost tab as default rather than just documents

src/shared/common/components/side-bar.jsx
```diff
@@ -59,19 +64,54 @@ class SideBar extends Component {
...
+        if (caseTabs.includes('people')) {
+            return 'PEOPLE';
+        }
+        if (caseTabs.includes('actions')) {
+            return 'CASE_ACTIONS';
+        }
+        if (caseTabs.includes('ex_gratia')) {
+            return 'EX_GRATIA';
+        }
+    }
```
robertdeniszczyc2, line 101: I think there is an opportunity here to make the order defined by config with a sort order, then here we take the first item depending on the sort order, rather than relying on the order hard coded in a method here.

Happy to talk through,

Cheers,
Rob

robertdeniszczyc2 (changes requested):

robertdeniszczyc2 (approved):

robertdeniszczyc2 (approved):

PR #970, opened 2022-05-11
HOCS-2083: change usage of Panel

LiamMacP:
Multiple usages of the Panel for actions pass the title as the summary
which breaks the Accessibility guidelines. All actions that pass title
incorrectly have been changed.

Also, includes a change that if the summary is not present then the
`div` does not present as an empty element.

### DIFF
--- a/server/services/action.js
+++ b/server/services/action.js
@@ -96,7 +96,7 @@ const actions = {
                     case actionTypes.CREATE_CASE: {
                         const { data: documentTags } = await getDocumentTags(context);
                         response = await createCase('/case', { caseType: context, form }, documentTags[0], headers);
-                        clientResponse = { summary: 'Case Created ', link: `${response.data.reference}` };
+                        clientResponse = { title: 'Case Created', child: { link: `${response.data.reference}` } };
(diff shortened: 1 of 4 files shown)


server/services/action.js
```diff
@@ -112,7 +112,7 @@ const actions = {
                     case actionTypes.BULK_CREATE_CASE: {
                         const { data: documentTags } = await getDocumentTags(context);
                         response = await createCase('/case/bulk', { caseType: context, form }, documentTags[0], headers);
-                        clientResponse = { summary: `Created ${response.data.count} new case${response.data.count > 1 ? 's' : ''}` };
+                        clientResponse = { child: { summary: `Created ${response.data.count} new case${response.data.count > 1 ? 's' : ''}` } };
```
gary-newbolt, line 115: should `summary` on this line be changed to `title` here too? And have no `child` object, or an empty one? That would seem more consistent with the other changes I think

LiamMacP: Good catch there - I've added better handling of nonexistant child in the Panel.

gary-newbolt (approved):
looks good

PR #976, opened 2022-05-24
HOCS-5097 Ensure required data present to render tab

robertdeniszczyc2:
Adds a check for existing stage data available in props
before rendering the tab. If the data doesn't exist, fetch
the stage data before then fetching and rendering the form.


LiamMacP:
Is there any tests we can add for this?

robertdeniszczyc2:
This is now abandoned as fixed by https://github.com/UKHomeOffice/hocs-frontend/pull/975

PR #1000, opened 2022-06-22
HOCS-5184: rework field processing

LiamMacP:
This change reworks the processing to allow for validation rules to be
applied before the stripping of case non-acceptable field types.

This involves effectively merging the validation step into the field
processing as they are tightly coupled. This then allows for the removal
of the data properties before forms are submitted to casework.


server/middleware/form/validation.js
```diff
@@ -605,9 +575,60 @@ function validationMiddleware(req, res, next) {
             }
         }
     }
+
+    /**
+     * Check to see if field component is of display only (without validation rules).
+     * @param field the field to check
+     * @returns {boolean} whether the field should be kept (true)
+     */
+    function isDisplayComponent(field) {
```
robertdeniszczyc2, line 584: We also have this logic in process.js on line 86, potential to share across with something like isComponentOfType(type) instead maybe?

```
return field.component === type
```

LiamMacP: I've made a change - good shout - think I've done it as you suggested.

robertdeniszczyc2 (approved):

PR #1022, opened 2022-07-05
HOCS-5274: search and profile configuration

LiamMacP:
Creation of static tenant configuration in a JSON file. This allows for
changes without backend database changes being required to change
frontend UI toggles.

Creation of search field JSON configuration file. This allows for all UI
logic for various case types to reside within the frontend and be
easily changeable without relying on an info service configuration call.

The search field generation code has also been refactored to remove a
redundant call and be more readable.

### DIFF
--- a/kube/kd/deployment.yaml
+++ b/kube/kd/deployment.yaml
@@ -353,6 +353,8 @@ spec:
           {{ end }}
           - name: SHOW_STACKTRACE_ON_ERROR_PAGE
             value: '{{.SHOW_ERROR_STACKTRACE}}'
+          - name: TENANT
+            value: '{{.DOMAIN}}'
         resources:
           limits:
             cpu: 1600m
--- a/server/config/searchFields/config.json
+++ b/server/config/searchFields/config.json
@@ -1,0 +1,811 @@
+{
+  "BF": [
+    {
+      "validation": [],
+      "props": {
+        "label": "Correspondent full name"
+      },
+      "name": "correspondent",
+      "component": "text"
+    },
+    {
+      "validation": [],
+      "props": {
+        "label": "Correspondent postcode"
+      },
+      "name": "correspondentPostcode",
+      "component": "text"
+    },
+    {
+      "validation": [],
+      "props": {
+        "label": "Correspondent email address"
+      },
+      "name": "correspondentEmail",
+      "component": "text"
+    },
+    {
+      "validation": [
+        "isValidDate",
+        "isBeforeToday"
+      ],
+      "props": {
+        "label": "Complainant date of birth"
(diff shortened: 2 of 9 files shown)


server/tenantConfig.js
```diff
@@ -1,31 +1,30 @@
-const listService = require('./services/list');
 const { isNotProd } = require('./config');
-const { v4: uuid } = require('uuid');
+const { fetchTenantConfiguration } = require('./config/tenant/tenantConfiguration');
+
+const profile = fetchTenantConfiguration(process.env.PROFILE);
```
LiamMacP, line 4: Should probably be renamed to `tenantConfig`.

server/config/tenant/config.json
```diff
@@ -0,0 +1,16 @@
+{
```
LiamMacP, line 1: This doesn't contain all of the profile - upon looking at it - I'm not sure that workstack columns (and a few others) should be part of this...

kube/kd/deployment.yaml
```diff
@@ -353,6 +353,8 @@ spec:
           {{ end }}
           - name: SHOW_STACKTRACE_ON_ERROR_PAGE
             value: '{{.SHOW_ERROR_STACKTRACE}}'
+          - name: PROFILE
+            value: '{{.DOMAIN}}
```
LiamMacP, line 357: This needs a closing quote.

server/config/tenant/config.json
LiamMacP: There is a ticket to have workstack columns in hocs-casework, so I guess I'm right.

server/config/tenant/config.json
```diff
@@ -0,0 +1,16 @@
...
+    "cs": {
+      "displayName": "Correspondence System",
+      "bulkCreate": true,
+      "deadlines": true,
+      "autoCreateAndAllocate": false,
```
nickmcmahon01, line 6: Who is this true for?

LiamMacP: No-one by the looks of it. I'll have a look to see if it's used in the codebase tomorrow.

server/config/searchFields/config.json
```diff
@@ -0,0 +1,811 @@
+{
+  "BF": [
+    {
+      "validation": [],
+      "props": {
```
SWhitlam, line 5: I think we could remove props and also consider renaming label to something like "description" or "title"?

LiamMacP: I think this should be a second pass? We typically have 'label' outside of props

SWhitlam (approved):

PR #1023, opened 2022-07-11
HOCS-5335: disable auto allocate

LiamMacP:
Initially when WCS was created autoCreateAndAllocate was set to true.
This was subsequently disabled but not respected in this config file -
this change deactivates this for the tenant.


SWhitlam (approved):

LiamMacP:
Changed to main - as 1.26.0 is head of main.

PR #1042, opened 2022-07-22
downgrade axios to 0.26.1 to resolve broken deployment

SWhitlam:


zuzak (approved):
Looks equivalent to 889a8b6d8a8f292d5d4f03be3bfdd22d98a3c38e.
Would be better with a properly cased commit message and a ticket number, but as this doesn't need QA the latter is probably less needed.

PR #1059, opened 2022-08-01
HOCS-4270: fix falsey casenote request

LiamMacP:
When sending a casenote with a falsey value, the codepath calls next
with an error but does not return at that point, this means that after
completion of the response a request to casework happens with a invalid
casenote. This is causing exceptions to be thrown from casework as we
don't support empty case notes and also express router failures.

This change fixes this by returning the value of the next as to
terminate the function.

### DIFF
--- a/server/middleware/__tests__/case.spec.js
+++ b/server/middleware/__tests__/case.spec.js
@@ -419,6 +419,7 @@ describe('Case middleware', () => {
             await createCaseNote(req, res, next);
             expect(next).toHaveBeenCalled();
             expect(res.locals.error).toEqual('Case note must not be blank');
+            expect(caseworkService.post.mock.calls.length).toEqual(0);
         });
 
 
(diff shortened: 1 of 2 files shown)


zuzak (approved):
A bit strange that we're doing `next(new Error('text'))` sometimes and `res.locals.error = 'text'; next()` other times?

zuzak:
Ideally your commit message would be an instruction ("Fix falsey casenote request") not a description ("Falsey casenote fix").

LiamMacP:
> A bit strange that we're doing `next(new Error('text'))` sometimes and `res.locals.error = 'text'; next()` other times?

Looks like we have a mix of returning objects vs expecting text. Multiple implementations because we don't have a 'standard'.

zuzak:
Oh GitHub is complaining about your email address -- check that before merging

PR #1064, opened 2022-08-04
HOCS-4781: Add POGR complaint entities

JosephRogerson:


LiamMacP (approved):
Looks good to me!

It would be beneficial to think about how we could simplify this. We have many tens of these now - is there a way we can just have a generic get entity list method and store the results in a map. 

I'm not overly convinced there is a benefit to loading these at startup - we could cache them after the first call for example (which removes a potential dependency on info before this can spin up). Out of scope for this ticket but maybe it will be beneficial to create this and flesh out? 

You will need to take care that data gets merged before this otherwise the frontend likely won't spin up.

PR #1075, opened 2022-08-12
HOCS-5361: rework somu implementation

LiamMacP:
SOMU form implementation currently runs through the actions mechanic for
retrieving the forms and handling the submission. While this was okay
for the initial development, as tenants are requiring specific
functionality this has become clunky and a pain to work around.

This change reworks SOMU implementations so that forms are retrieved
like 'global' forms outside of a workflow. This gives us the ability to
tailor them to the tenants themselves. Additionally, these forms can
also utilise normal case data attributes for visibility conditions.

This change also cleans up some of the initial SOMU development in the
following aspects:
- All posts go through the same API handling
- Getting an individual SOMU item no longer first retrieves them all
- Cleans up submission to remove unused code

Support the ability to move back to `/case/<UUID>/stage/<UUID>` if
action is undefined and UUIDs exist in request parameters. This is
useful for dynamic backlink components that are served from the database
that should not progress workflows but return to the case.


server/services/form.js
```diff
@@ -322,14 +333,49 @@ const getGlobalFormForCase = async (req, res, next) => {
...
+        params: { action, caseId },
+        somuType: { uuid, schema: { forms = {} } = {} } = {},
+        somuItemData } = req;
+    const logger = getLogger(requestId);
+
+    try {
+        const formName = forms[action];
+
+        if (!formName) {
+            return next(new FormServiceError(`Schema does not contain contain form definition for Somu Type: ${uuid}`));
```
robertdeniszczyc2, line 348:
```suggestion
            return next(new FormServiceError(`Schema does not contain form definition for Somu Type: ${uuid}`));
```

server/services/form.js
```diff
@@ -322,14 +333,49 @@ const getGlobalFormForCase = async (req, res, next) => {
...
+
+        if (somuItemData) {
+            form.data = { ...form.data, ...somuItemData };
+        }
+
+        req.form = form;
+
+        next();
+    } catch (error) {
+        logger.error('CASEOWKR_SOMU_FORM_FAILURE', { message: error.message, stack: error.stack });
```
robertdeniszczyc2, line 364:
```suggestion
        logger.error('CASEWORK_SOMU_FORM_FAILURE', { message: error.message, stack: error.stack });
```

PR #1082, opened 2022-08-15
[HOCS-4838] Run on Node 16.x

robertdeniszczyc2:
This change allows the enables the module to build and run on Node version 16.x (current LTS).

Also included:
- package-lock.json uses lockfileVersion 2
- Fixed ESLint warnings, and marked several false positives as TODO
- README updated


nickmcmahon01:
🥇

robertdeniszczyc2:
The diff for the package-lock.json file is so big because it's now using lockfile version 2 rather than 1, and the structure of the lockfile has changed.

LiamMacP (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14-alpine as base
+FROM node:16.16.0-alpine3.15 as base
```
nickmcmahon01, line 1: can this be node:16-alpine please

nickmcmahon01 (dismissed):

Dockerfile
nickmcmahon01, line 1:
```suggestion
FROM node:16-alpine as base
```

PR #1093, opened 2022-08-17
HOCS-5483: Remove escalation document requirement

JosephRogerson:


server/services/forms/index.js
```diff
@@ -128,6 +128,10 @@ const formDefinitions = {
                     action: CREATE_CASE,
                     next: {
                         action: 'CONFIRMATION_SUMMARY'
+                    },
+                    customConfig: {
```
LiamMacP, line 132: Can we call this config? Any additional config is custom...

server/services/forms/schemas/document-add-next-case.js
```diff
@@ -5,6 +5,20 @@ const { MIN_ALLOWABLE_YEAR } = require('../../../libs/dateHelpers');
...
+    const documentRequired = options.customConfig ? options.customConfig.documentRequired : true;
+    const documentLabel = options.customConfig ? options.customConfig.documentLabel : 'There must be a least 1 document to upload.';
+
+    const documentComponent = Component('add-document', 'add_document')
+        .withValidator('hasWhitelistedExtension')
+        .withValidator('fileLimit')
+        .withProp('label', documentLabel)
+        .withProp('hint', `There is a limit of ${DOCUMENT_BULK_LIMIT} documents`)
+        .withProp('allowMultiple', true)
+        .withProp('whitelist', 'DOCUMENT_EXTENSION_WHITELIST')
```
LiamMacP, line 17:
```suggestion
        .withProp('whitelist', 'DOCUMENT_EXTENSION_WHITELIST');
```

server/services/forms/schemas/document-add-next-case.js
```diff
@@ -5,6 +5,20 @@ const { MIN_ALLOWABLE_YEAR } = require('../../../libs/dateHelpers');
 
 module.exports = (options) => {
     const fromCaseUUID = options?.data?.from ? options.data.from : '';
+    const documentRequired = options.customConfig ? options.customConfig.documentRequired : true;
```
LiamMacP, line 8:
```suggestion
    const documentRequired = options.config?.documentRequired || true;
```

server/services/forms/index.js
LiamMacP, line 132:
```suggestion
                    config: {
```

server/services/forms/index.js
```diff
@@ -383,8 +387,9 @@ module.exports = {
                 } else {
                     formDefinition = formDefinitions[context.toUpperCase()][workflow.toUpperCase()][action.toUpperCase()];
                 }
+                let customConfig = formDefinition.customConfig;
 
-                let form = await formDefinition.builder.call(this, { data });
+                let form = await formDefinition.builder.call(this, { data, customConfig });
```
LiamMacP, line 392: You can just expand this out? `{ data, config: formDefinition.config}`
Or maybe even `{ data, formDefinition.config}`

server/services/forms/schemas/document-add-next-case.js
```diff
@@ -5,6 +5,20 @@ const { MIN_ALLOWABLE_YEAR } = require('../../../libs/dateHelpers');
...
+    const documentComponent = Component('add-document', 'add_document')
+        .withValidator('hasWhitelistedExtension')
+        .withValidator('fileLimit')
+        .withProp('label', documentLabel)
+        .withProp('hint', `There is a limit of ${DOCUMENT_BULK_LIMIT} documents`)
+        .withProp('allowMultiple', true)
+        .withProp('whitelist', 'DOCUMENT_EXTENSION_WHITELIST')
+
+    if (documentRequired) {
+        documentComponent.withValidator('required')
```
LiamMacP, line 20:
```suggestion
        documentComponent.withValidator('required');
```

server/services/forms/schemas/document-add-next-case.js
LiamMacP, line 8: Unsure from a readability point of view but you could do: 

`options?.config?.documentRequired || true`

JosephRogerson: I don't think that will work because when `documentRequired` is set to false that will output true

LiamMacP: Thats true - look into the null coelescing operator `??`, that only checks for undefined or null.

LiamMacP (approved):

PR #1104, opened 2022-08-26
HOCS-4794 Update GOVUK Frontend to 4.3.1

robertdeniszczyc2:
Updates GOVUK Frontend package to 4.3.1 from 3.14.0, and makes required changes for affected elements between the versions.

#### Accordion
The design of the Accordion component changed in 4.0.0. These changes are reflected here and markup changes also applied.

#### Header
The HTML for the header was updated in 4.0.0 for better screen reader support.

#### Hint text
Hint text elements were switched to divs from spans in 3.8.0, 4.0.0 introduced a breaking change were the styles were removed for the spans. The hint text spans have been converted to divs.

#### Tabs
The design of the sidebar tabs have been updated to reflect the latest tab component design.

#### Skip link
The skip link component HTML has been updated in line with changes introduced in 4.0.0.

#### Error messages
Error messages now use `p` tags rather than `span`

#### Change date input from type="number" to inputmode="numeric"
This was a Recommended change in 4.0.0 following research by the Design System team. In 4.1.0 a Recommended change was made to remove the pattern from the element.

#### Footer
Fix fallback logo so Chrome will not flag it to screen readers.


src/shared/common/forms/accordion.jsx
```diff
@@ -25,15 +26,36 @@ function Section({ data, errors, index, items, title, updateState, page }) {
     const clickHandler = React.useCallback(event => {
         event.preventDefault();
         setVisible(!isVisible);
-    }, [isVisible]);
+        if (toggleText === 'Show') {
```
zuzak, line 29: Is there a more reliable way to determine state than relying on the text? Can we use `isVisible`?

server/layout/html.js
```diff
@@ -41,7 +41,7 @@ const render = ({
 </head>
 
 <body class="govuk-template__body ">
-
+    <script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
```
zuzak, line 44: You could make this shorter if you wanted:
```suggestion
    <script>document.body.className = (document.body.className ? document.body.className : '') + ' js-enabled';</script>
```

src/shared/common/forms/accordion.jsx
```diff
@@ -12,6 +12,7 @@ function Section({ data, errors, index, items, title, updateState, page }) {
     const sectionHasValidationError = errors && Array.isArray(items) && items.some(itemHasValidationError(errors));
     const [isVisible, setVisible] = useState(sectionHasValidationError);
     const [isFocussed, setFocussed] = useState(false);
+    const [toggleText, setToggleText] = useState('Show');
```
zuzak, line 15: You're switching between boolean and string here, which could cause confusion.

zuzak (approved):

src/shared/common/forms/accordion.jsx
robertdeniszczyc2: Where is it setting as boolean sorry I can't spot it? Here it's initialising toggleText as string 'Show' and further down toggleText is set to string 'Hide.

zuzak: ah, I am looking at `useState(false)` and misreading line 14.

PR #1106, opened 2022-09-01
HOCS-5517: enquiry reason and enquiry subject display simple name fix

AlexW013:


server/lists/adapters/case-summary.js
```diff
@@ -34,6 +34,22 @@ const createAdditionalFields = async (additionalFields = [], fetchList) => {
...
 
+const fetchChoices = async (choices, fetchList) => {
+    if (choices.includes(', ')) {
+        const choicesList = [];
+        const listNames = choices.split(', ');
+        for (var j = 0; j < listNames.length; j++) {
+            const fetchedList = await fetchList(listNames[j]);
+            for (var k = 0; k < fetchedList.length; k++) {
+                choicesList.push(fetchedList[k]);
+            }
```
zuzak, line 45: There's a few native functions in `Array.prototype` that might be useful to make more robust code that doesn't require a for-loop (and might be faster).

If you want to combine two arrays, the [`Array.prototype.concat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) function might be better than a for loop. If you have an array of arrays, you can flatten it into a single array with `.flat()`.

Similarly, if you are wanting to perform an operation on every element in an array, [`Array.prototype.map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) might be a more readable choice.


(If you really wanted -- I wouldn't recommend this? -- you could collapse it all into a one-liner:)
```js
const choicesList = choices.split(', ').map((item) => await fetchList(item)).flat()
```

server/lists/adapters/case-summary.js
```diff
@@ -34,6 +34,22 @@ const createAdditionalFields = async (additionalFields = [], fetchList) => {
     return results;
 };
 
+const fetchChoices = async (choices, fetchList) => {
+    if (choices.includes(', ')) {
```
zuzak, line 38: If the search string isn't found, Array.prototype.split() will still return an array (with one element), sp you don't need a special-case for if `choices` doesn't include your search string. If you wanted, you could remove this `if` statement.

```js
"foo".split(', ')      // [ "foo" ]
"foo, bar".split(', ') // [ "foo", "bar" ]
```

LiamMacP:
Closed as fixed under HOCS-5545.

PR #1109, opened 2022-09-05
HOCS-5560: rework unallocated accordion

LiamMacP:
Rework of the unallocated accordion to support showing `somu-list`s.
This works by first analyzing the component and if it is a `somu-list`
it gets the meta-data and then the case items for it. This populates the
custom choice behaviour and then composes the titles for each of the
items.

The data for this unallocated accordion is calculated first which is
then used to filter out fields from the accordion that do not have a
data mapping.

### DIFF
--- a/server/libs/objectHelpers.js
+++ b/server/libs/objectHelpers.js
@@ -1,0 +1,14 @@
+/**
+ * @param object a object with a single value
+ * @returns {array} array with index 0 being the name and index 1 being the value
+ */
+const getObjectNameValue = (object) => {
+    if (object &&
+        Object.keys(object).length >= 1 &&
+        Object.getPrototypeOf(object) === Object.prototype) {
+        return Object.entries(object)[0];
+    }
+    return ['',''];
+};
+
+module.exports = getObjectNameValue;
--- a/server/lists/adapters/case-view-all-stages.js
+++ b/server/lists/adapters/case-view-all-stages.js
@@ -155,7 +155,6 @@ const hydrateFields = async (fieldTemplate, template, name, request) => {
         if (choices && typeof choices === 'object') {
             for (const [name, choice] of Object.entries(choices)) {
                 if (choice) {
-                    //to
                     Object.assign(choiceObj, { [name]: await request.fetchList(choice) });
                 }
             }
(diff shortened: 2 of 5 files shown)


server/libs/objectHelpers.js
```diff
@@ -0,0 +1,14 @@
+/**
```
LiamMacP, line 1: I don't like how this file is a duplicate from the `src` folder - we really need to look into a proper way of sharing files.

server/lists/adapters/case-view-all-stages.js
```diff
@@ -38,103 +34,79 @@ const getRequestStatus = (dueDate, { status, decision }) => {
...
-                fieldTemplate => getComponentFromField(fieldTemplate, template))
-                .filter(component => component !== undefined); // remove empty elements caused by hidden fields .etc
-            const stageName = await fromStaticList('S_STAGETYPES', stageId);
-
-            return { title: stageName, items: stageFields };
-        }))).filter(stage => {
-            return stage.items && stage.items.length > 0;
-        }); // filter out empty sections
-
     const data = (await Promise.all(Object.entries(template.schema.fields)
```
LiamMacP, line 104: This loops around all of the fields regardless of if they are shown or not - resulting in potentially a high cost value for this method. If the same field is present multiple times, this would also cause the same impact. This should be looked at as part of the unallocated accordion rewrite. HOCS-5569

server/libs/objectHelpers.js
robertdeniszczyc2: Potentially could turn it into a Package, but for a very specific use case...

server/lists/adapters/case-view-all-stages.js
```diff
@@ -157,45 +143,54 @@ module.exports = async (template, { fromStaticList }) => {
...
+    }
+
+    if (fieldTemplate.component === 'somu-list') {
+        const { somuType, choices } = fieldTemplate.props;
+        const choiceObj = {};
+
+        if (choices && typeof choices === 'object') {
+            for (const [name, choice] of Object.entries(choices)) {
+                if (choice) {
+                    //to
```
robertdeniszczyc2, line 158:
```suggestion
```

robertdeniszczyc2 (approved):

PR #1111, opened 2022-09-06
HOCS-4974 Update GOVUK Frontend to 4.3.1

robertdeniszczyc2:
Updates GOVUK Frontend package to 4.3.1 from 3.14.0, and makes required changes for affected elements between the versions.

### Impacted areas
- Accordion
The design of the Accordion component changed in 4.0.0. These changes are reflected here and markup changes also applied.

- Header
The HTML for the header was updated in 4.0.0 for better screen reader support.

- Hint text
Hint text elements were switched to divs from spans in 3.8.0, 4.0.0 introduced a breaking change were the styles were removed for the spans. The hint text spans have been converted to divs.

- Tabs
The design of the sidebar tabs have been updated to reflect the latest tab component design.

- Skip link
The skip link component HTML has been updated in line with changes introduced in 4.0.0.

- Error messages
Error messages now use p tags rather than span

- Change date input from type="number" to inputmode="numeric"
This was a Recommended change in 4.0.0 following research by the Design System team. In 4.1.0 a Recommended change was made to remove the pattern from the element.

- Footer
Fix fallback logo so Chrome will not flag it to screen readers.

### Also includes:
- Removes govuk-frontend JS, fixes components which relied on it.
- Fixes radio button groups which relied on GOVUK Frontend JS.
- Fixes issues with Accordion component for toggle text behaviour.


src/shared/common/forms/radio-group.jsx
```diff
@@ -14,11 +14,6 @@ class Radio extends Component {
 
     componentDidMount() {
         this.props.updateState({ [this.props.name]: this.props.value });
-        // eslint-disable-next-line no-undef
-        if (window.GOVUKFrontend) {
```
LiamMacP, line 18: I'm glad to see the back of this.

LiamMacP (approved):
The changes seem fine - can we put this on QAX and ask QA to run a regression before merge please?

PR #1114, opened 2022-09-07
HOCS-5467 - Fix to display - user entered fields to the user in People tab

vinothon84:
HOCS-5467 - Fix to display - user entered fields to the user in 
People tab

Previously, we only showed the address field in the table if address1
was populated. There was a bug identified where if the user don't fill 
address1 but fill rest of the field like address2, town, postcode or 

This commit changes the behaviour such that when user fills any one 
of the filed as above - it will be shown to the user.

### DIFF
--- a/src/shared/common/components/people.jsx
+++ b/src/shared/common/components/people.jsx
@@ -64,15 +64,16 @@ class People extends Component {
                             <td className='govuk-table__cell'>{person.organisation}</td>
                         </tr>
                         }
-                        {person.address.address1 &&
+                        {(person.address.address1 || person.address.address2 || person.address.address3
+                                || person.address.postcode || person.address.country ) &&
                         <tr className='govuk-table__cell'>
                             <th className='govuk-table__header padding-left--small govuk-!-width-one-third'>Address</th>
                             <td className='govuk-table__cell'>
-                                {person.address.address1 && <> <span>{person.address.address1}</span> </>}
-                                {person.address.address2 && <> <br/> <span>{person.address.address2}</span> </>}
-                                {person.address.address3 && <> <br/> <span>{person.address.address3}</span> </>}
-                                {person.address.postcode && <> <br/> <span>{person.address.postcode}</span> </>}
-                                {person.address.country && <> <br/> <span>{person.address.country}</span> </>}
+                                {person.address.address1 && <> <span>{person.address.address1} </span> <br/></>}
+                                {person.address.address2 && <> <span>{person.address.address2} </span> <br/></>}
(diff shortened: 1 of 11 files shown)


src/shared/common/components/people.jsx
```diff
@@ -64,15 +64,16 @@ class People extends Component {
...
-                                {person.address.address1 && <> <span>{person.address.address1}</span> </>}
-                                {person.address.address2 && <> <br/> <span>{person.address.address2}</span> </>}
-                                {person.address.address3 && <> <br/> <span>{person.address.address3}</span> </>}
-                                {person.address.postcode && <> <br/> <span>{person.address.postcode}</span> </>}
-                                {person.address.country && <> <br/> <span>{person.address.country}</span> </>}
+                                {person.address.address1 && <> <span>{person.address.address1} </span> <br/></>}
+                                {person.address.address2 && <> <span>{person.address.address2} </span> <br/></>}
+                                {person.address.address3 && <> <span>{person.address.address3} </span> <br/></>}
+                                {person.address.postcode && <> <span>{person.address.postcode} </span> <br/> </>}
+                                {person.address.country && <>  <span>{person.address.country} </span> <br/> </>}
```
LiamMacP, line 76: You have revered the <br/> - is the final one required here?

src/shared/common/components/people.jsx
```diff
@@ -64,15 +64,16 @@ class People extends Component {
                             <td className='govuk-table__cell'>{person.organisation}</td>
                         </tr>
                         }
-                        {person.address.address1 &&
+                        {(person.address.address1 || person.address.address2 || person.address.address3
```
LiamMacP, line 67: This is a change to functionality - should we be adding a test for this component to cover this?

src/shared/common/components/people.jsx
zuzak, line 76: Instead of `<br>` perhaps we should use a block-level element (e.g. `<div>` instead of `<span>`, or use the `white-space` CSS property? As written, you'll always have an unnecessary terminating line break -- although it won't affect presentation as the browser will ignore it when computing whitespace.

Incidentally, in HTML5 you can just write `<br>`; although `<br/>` is valid (it's for backwards compatibility).

LiamMacP: Good information there - no action here - keeping open for information purposes.

vinothon84: I will remove the last <br/> as its redundant here.

LiamMacP (approved):

PR #1125, opened 2022-09-29
HOCS-5629: Fix IEDET final response upload

JosephRogerson:


server/lists/index.js
```diff
@@ -624,6 +624,11 @@ module.exports = {
             endpoint: '/case/document/reference/${caseId}/?type=Draft',
             adapter: documentListAdapter
         },
+        CASE_DOCUMENT_LIST_IEDET_FINAL_RESPONSE: {
```
LiamMacP, line 627: We have generic types like `Final response` across the system. Any chance you could create a ticket to attempt to simplify this? We really need a list of tags across the system and then cases just opt in. Then we can just have `CASE_DOCUMENT_FINAL_RESPONSE` list rather than loads.

LiamMacP (approved):

server/lists/index.js
JosephRogerson: Yeah, I'll create a ticket for this. There's 4 different kinds of draft in there already so we should definitely have a look at simplifying this

PR #1126, opened 2022-10-04
HOCS-5536: Enable visibilityConditions on mapped fields

JosephRogerson:
This change allows visibility conditions to be passed as a prop to mapped fields


LiamMacP (approved):

LiamMacP:
I think this is fine - I'm unsure about if its required. The default unallocated view just wouldn't show the data if it wasn't in the map.

PR #1130, opened 2022-10-05
HOCS-5632: IEDET - Complaint sub categories are not displayed correctly in Case Details accordion

AlexW013:


src/shared/common/forms/mapped-display.jsx
```diff
@@ -17,6 +17,19 @@ const renderCheckbox = (label, showLabel, choicesLabel) =>  {
     );
 };
 
+const renderCheckboxGroup = (mappings, value, label) => {
+    let displayValue;
```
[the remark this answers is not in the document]
AlexW013: I have simplified down to one line but I don't think replace() can be used since it only replaces the first ',' value it finds when could be multiple that need replacing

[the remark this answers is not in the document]
AlexW013: Actually simplified way further

src/shared/common/forms/mapped-display.jsx
```diff
@@ -36,7 +36,7 @@ class MappedDisplay extends Component {
             return (renderCheckbox(label, this.props.showLabel, choices[0].label));
         } else {
             return (
-                <span className='govuk-body full-width'><strong>{label}: </strong>{mappings.get(value) ? mappings.get(value) : value}</span>
+                <span className='govuk-body full-width'><strong>{label}: </strong>{mappings.get(value) ? mappings.get(value) : value.split(',').join(', ')}</span>
```
LiamMacP, line 39: You could use replaceAll here instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll

My only thing is I believe the split and join is O(n^2) performance whereas I'd presume this is O(n).

src/shared/common/forms/mapped-display.jsx
```diff
@@ -36,7 +36,7 @@ class MappedDisplay extends Component {
             return (renderCheckbox(label, this.props.showLabel, choices[0].label));
         } else {
             return (
-                <span className='govuk-body full-width'><strong>{label}: </strong>{mappings.get(value) ? mappings.get(value) : value}</span>
+                <span className='govuk-body full-width'><strong>{label}: </strong>{mappings.get(value) ? mappings.get(value) : value.replaceAll(',', ', ')}</span>
```
SWhitlam, line 39: Should we be more specific in the else that it's rendering a checkbox-group?

AlexW013: Yeah I get what you're saying. It might not be a checkbox group though, it could be any component that isn't specifically a checkbox. I did have a version of this where it called a method to render a checkbox group but the only difference in that method was the addition off the replaceAll so I decided to just combine it here into line 39

SWhitlam (approved):

PR #1160, opened 2022-10-19
Nm 19 spike

nickmcmahon01:


.github/workflows/test.yml
```diff
@@ -7,4 +7,4 @@ jobs:
   test:
     uses: UKHomeOffice/hocs-github-actions/.github/workflows/test-npm.yml@v2
     with:
-      node-version-matrix: '[ "16.x" , "18.x" ]'
+      node-version-matrix: '[ "18.x" , "19.x" ]'
```
LiamMacP, line 10: We're waiting till next week to move to 18, might be worth 16->19 if theres compatability.

nickmcmahon01: we can block PRs on failing code to catch issues for 20 early though, and it looks like is passes currently.

LiamMacP: Even better if it passed - think 16, 18 and 19 in the matrix is valuable.

PR #1183, opened 2022-11-03
HOCS-2005: rework Axios error handling

LiamMacP:
When an error is thrown within the express middleware there were a lot of instances of logging the entire Axios object. This isn't useful, and ends up spamming our SIEM.

This change removes this instances and adds a axios error middleware that transposes this error into a `GenericError`.

This change also (partially?) fixes an `Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client` exception that we we're having.

### DIFF
--- a/server/middleware/request.js
+++ b/server/middleware/request.js
@@ -1,12 +1,22 @@
 const logger = require('../libs/logger');
-const { ValidationError } = require('../models/error');
+const { GenericError, ValidationError } = require('../models/error');
 const { isProduction } = require('../config');
 const { v4: uuid } = require('uuid');
 const listService = require('../services/list/');
 
-/* eslint-disable-next-line  no-unused-vars*/
-function apiErrorMiddleware(err, req, res, next) {
+function axiosErrorMiddleware(err, _req, _res, next) {
+    if (err.isAxiosError) {
+        if (err.response) {
+            return next(new GenericError(err.response.data, err.response.status));
+        } else if (err.request) {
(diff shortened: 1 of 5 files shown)


server/middleware/request.js
```diff
@@ -1,12 +1,22 @@
 const logger = require('../libs/logger');
-const { ValidationError } = require('../models/error');
+const { GenericError, ValidationError } = require('../models/error');
 const { isProduction } = require('../config');
 const { v4: uuid } = require('uuid');
 const listService = require('../services/list/');
 
-/* eslint-disable-next-line  no-unused-vars*/
-function apiErrorMiddleware(err, req, res, next) {
+function axiosErrorMiddleware(err, _req, _res, next) {
```
robertdeniszczyc2, line 7: We could potentially add tests for this in request.spec.js

server/middleware/stage.js
```diff
@@ -30,48 +27,42 @@ async function stageApiResponseMiddleware(req, res, next) {
...
             userUUID: user.uuid,
         }, { headers: User.createHeaders(user) });
     } catch (error) {
-        logger.error(error);
-        next(error);
-    } finally {
-        next();
+        return next(error);
     }
+    next();
```
zuzak, line 65: Probably best to not return a function sometimes and undefined other times?
```suggestion
    return next();
```

zuzak: (the same for the other functions)

LiamMacP: Doesn't actually have an impact but I've done it to make it clear to anyone else in the future...

robertdeniszczyc2 (approved):

PR #1194, opened 2022-11-17
HOCS-5503: Change IEDET categories screen

JosephRogerson:


LiamMacP (commented):
It would be good if we could get some tests for these things?

JosephRogerson:
> It would be good if we could get some tests for these things?

Added tests

LiamMacP (approved):

PR #1205, opened 2022-12-01
HOCS-5566-Add tag row to show Rejected by PSU tag

vinothon84:
HOCS-5566 - workstack tag row been included in 
order to displaya tag for a case. This particular
PR would help to test Rejected by PSU tag for PSU
to IEDET rejected case.


src/shared/common/components/workstack.jsx
```diff
@@ -325,6 +325,7 @@ class WorkstackAllocate extends Component {
             case ColumnRenderer.CASE_LINK:
                 return <td key={row.uuid + column.dataValueKey} className='govuk-table__cell'>
                     <Link to={`/case/${row.caseUUID}/stage/${row.uuid}`} className='govuk-link govuk-!-margin-right-3'>{value}</Link>
+                    <Tags row={row} />
```
LiamMacP, line 328: This could do with a test if possible. Also as mentioned in the hocs-workflow PR would be good if we could improve this with a mapping change and allow for FE class colour. This might need to be done in the server side though...

vinothon84: I have added build changes for having a customised way of pulling name and class colour as part of the PR. 
Would you be able to review this again ?

server/config/tagType/tagConfig.json
```diff
@@ -0,0 +1,14 @@
...
+  "REJECTED_BY_PSU": [
+    {
+      "tagName": "REJECTED BY PSU",
+      "tagLabel": "govuk-tag govuk-tag--red"
+    }
+  ],
+  "HS": [
+    {
+      "tagName" : "HS",
+      "tagLabel": "govuk-tag govuk-!-margin-right-1"
```
LiamMacP, line 11: `govuk-!-margin-right-1`

Feels a little odd to have that on just this tag? I haven't pulled and checked the branch but this might be worth doing in the Tag component if theres more than one? Or we could bake some CSS in all apart from N-th last?

Happy to jump on a call for this 🙂

server/config/tagType/tagConfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "REJECTED_BY_PSU": [
```
LiamMacP, line 2: Is an array the correct thing here? Looks to me like theres only one item so can be an object?

server/config/tagType/tagType.js
```diff
@@ -0,0 +1,9 @@
+const config = require('./tagConfig.json');
+
+const fetchTagType = (name) => {
+    return config[name] ?? [];
```
LiamMacP, line 4: As this is a direct lookup per tag, this probably should return an object and not an array?

src/shared/common/components/workstack-tag.jsx
```diff
@@ -1,12 +1,22 @@
...
 
 export default function Tags(props) {
     let tags = [];
 
     if (props.row.tag && props.row.tag.length > 0) {
         tags = props.row.tag.map(tag => {
-            return <span key={tag} className="govuk-tag govuk-!-margin-right-1">{tag}</span>;
+            const tagDetails = fetchTagType([tag]);
+            let tagName = "DEFAULT";
+            let tagLabel= "govuk-tag govuk-!-margin-right-1";
```
LiamMacP, line 12: My gut feeling with these is that this should be returned within the tagType:L4 default object? I also think defaulting the name to the tag value e.g., `REJECTED_BY_PSU` would be more beneficial than `DEFAULT`?

LiamMacP (changes requested):
Good work on this! Some small changes highlighted - reach out if you need a hand.

server/config/tagType/tagConfig.json
vinothon84: Agreed and will incorporate this.

src/shared/common/components/workstack-tag.jsx
vinothon84: My thought process was like if any requirement for adding a new tag comes in future then DEV will add this in hocs-workflow but the corresponding expected data would be missing in hocs-frontend config which would cause frontend to fail instead if we have this configuration UI would display DEFAULT so that DEV would have a chance to fix this by adding it to config.

Let's discuss in today's call.

vinothon84: Thanks for your feedback and now code looks nice & tidy.

src/shared/common/components/__tests__/workstack-tag.spec.jsx
```diff
@@ -18,6 +18,6 @@ describe('Workstack Tag component', () => {
                 tag: ['HS']
             }
         };
-        expect(Tags(props).length).toBe(1);
+        expect(Tags(props.row.tag)).toBeDefined();
```
LiamMacP, line 21: This is incorrect now? `row` is no longer passed in?

src/shared/common/components/workstack-tag.jsx
```diff
@@ -1,20 +1,23 @@
...
+                className={tagDetails?.displayClass ?? 'govuk-tag govuk-!-margin-right-1'}>{tagDetails?.label ?? tag}</span>;
         });
     }
 
     return tags;
 }
 
 Tags.propTypes = {
-    row: PropTypes.object,
     tag: PropTypes.array
```
LiamMacP, line 21: We could ensure a default prop here and further get rid of the tag exists check? 

Look up React defaultProps...

LiamMacP (changes requested):

src/shared/common/components/__tests__/workstack-tag.spec.jsx
vinothon84: I have updated the test cases now.

LiamMacP (approved):

PR #1209, opened 2022-12-05
HOCS-5564: Implement business events

JosephRogerson:


server/lists/adapters/case-notes.js
```diff
@@ -202,7 +202,10 @@ const typeAdaptors = {
     OFFLINE_CASE_TRANSFER : ({ note }) => ({
         note: note,
         title: 'Offline Case Transfer'
-    })
+    }),
+    DATA_FIELD_UPDATED: ({ body }) => ({
```
LiamMacP, line 206: If this is a changed value, we should say that the field has changed from <value> to <new value>?

JosephRogerson: Good question, this has come from the original ticket. I've asked Tom and we'll see what his opinion of changing it is

PR #1216, opened 2022-12-12
HOCS-5979: Add IEDET category validation

JosephRogerson:


server/middleware/form/process.js
```diff
@@ -110,8 +127,11 @@ function processMiddleware(req, res, next) {
             .filter(Boolean);
 
         const fieldData = visibleFields.reduce(createReducer(data, req), {});
+        const activeValidation = getActiveValidators(schema.validation, data);
+
+        console.log(activeValidation);
```
LiamMacP, line 132:
```suggestion
        console.log(activeValidation);
```

server/middleware/form/process.js
```diff
@@ -97,6 +97,23 @@ const isUnacceptedDataFields = (field) => {
         isFieldComponentOfType(field, 'somu-list');
 };
 
+function checkValidationCondition(validator, caseData) {
+    if (Object.prototype.hasOwnProperty.call(validator, 'condition') && caseData[validator.condition.fieldName] !== validator.condition.value) {
```
LiamMacP, line 101: Would it be simpler to inverse this? Return true only in this case and default to false? Unsure if that has side effects though...

server/middleware/form/process.js
```diff
@@ -97,6 +97,23 @@ const isUnacceptedDataFields = (field) => {
...
+        return false;
+    }
+    return true;
+}
+
+function getActiveValidators(validationSchemaJson, caseData) {
+    const validationSchema = typeof validationSchemaJson === 'string' ? JSON.parse(validationSchemaJson) : validationSchemaJson;
```
LiamMacP, line 108: Would a try...catch work well here, there is still the edge case that the string would not be valid JSON

server/middleware/form/process.js
```diff
@@ -97,6 +97,23 @@ const isUnacceptedDataFields = (field) => {
...
+    if (!validationSchema) {
+        return [];
+    }
+    const validationSchemaArray = Array.isArray(validationSchema) ? validationSchema : Array.of(validationSchema);
```
LiamMacP, line 112: Is the `Array.of` required? If it is an array is would just work?

LiamMacP (approved):

PR #1218, opened 2022-12-16
HOCS-2298: Load case error messages for non-real reference is poorly worded

AlexW013:


server/routes/dashboard.js
```diff
@@ -70,7 +70,7 @@ router.post('/api/search/reference', async (req, res) => {
                 return res.json({ forwardProps: { workstack: res.locals.workstack }, redirect: '/search/reference' });
             }
         }
-        res.json({ errors: { 'case-reference': 'No active workflows for case' } });
+        res.json({ errors: { 'case-reference': 'No cases have that reference - check what you have entered and try again' } });
```
robertdeniszczyc2, line 73: The wording for this is inaccurate. You could enter a case reference of a closed case and it wouldn't come up, but if you searched for the case with the case reference via Search, Search would return the closed case.

This needs putting back to Design to refine the wording.

robertdeniszczyc2 (changes requested):

robertdeniszczyc2 (approved):

PR #1226, opened 2023-01-09
⬆️ Update dependancies

robertdeniszczyc2:
Updates ESLint, Webpack, govuk-frontend, aws-sdk, various babel, autoprefixer, core-js, sass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -217,15 +217,26 @@
             "integrity": "sha512-Xni35NKzjgMrwevysHTCArtLDpPvye8zV/0E4EyYn43P7/7qvQwPh9BGkHewbMulVntbigmcT7rdX3BNo9wRJg=="
         },
         "node_modules/@aws-sdk/abort-controller": {
-            "version": "3.197.0",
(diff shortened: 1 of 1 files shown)


LiamMacP (approved):
Should this not also change the `package.json`? Or is it carated?

robertdeniszczyc2:
> Should this not also change the `package.json`? Or is it carated?

I updated them via `npm outdated` so it's just updated the lock file as it's still `package.json` file compliant. I can manually update the package.json file numbers if we want them to match, but changing the numbers there won't have a difference on the output.

For someone cloning the project from a clean slate and running `npm install` the versions in the `package-lock` take precedence so would be consistent across anyone using the project.

LiamMacP:
> > Should this not also change the `package.json`? Or is it carated?
> 
> I updated them via `npm outdated` so it's just updated the lock file as it's still `package.json` file compliant. I can manually update the package.json file numbers if we want them to match, but changing the numbers there won't have a difference on the output.
> 
> For someone cloning the project from a clean slate and running `npm install` the versions in the `package-lock` take precedence so would be consistent across anyone using the project.

I'm fine with this - just wanted to make sure 🙂

PR #1232, opened 2023-01-26
Chore: Migrate to Sas actions

nickmcmahon01:


.github/workflows/anchore-scan.yml
```diff
@@ -0,0 +1,18 @@
...
+  pull_request:
+    types: [ labeled, opened, reopened, synchronize ]
+  schedule:
+    - cron: '45 12 * * 1'
+
+jobs:
+  scan:
+    uses: UKHomeOffice/sas-github-workflows/.github/workflows/anchore-npm.yml@v1
+    with:
+      installCommand: 'ci --production=false --no-optional'
```
LiamMacP, line 15: Production false here but build as production?

nickmcmahon01: I see what you mean about bringing in dev dependencies (or not), but it's building a docker image and scanning the image not scanning the project itself. so if there's an issue it might be the docker file including dev dependencies when it doesn't need to?
it's using that same command to pull in deps and then runs build-prod here:
https://github.com/UKHomeOffice/hocs-github-actions/blob/137a415a92653f5ecd93ff8a62dbad3a4084c3c8/.github/workflows/docker-build-branch-npm.yml#L55

nickmcmahon01: so what this is doing is splitting that docker build-and-push action into seperate build and push steps and putting an anchore scan in between them, happy to 🤙

LiamMacP (approved):

PR #1237, opened 2023-02-07
HOCS-5615: rework Workstack form submission

LiamMacP:
The current use of FormData and `append` results in undefined values being passed with the value as string "undefined".

This is causing parsing errors within the Node server and further downstream as "undefined" is not valid JSON.

This change reworks this FormData propagation to strip out any underlying null/undefined.

### DIFF
--- a/src/shared/pages/workstack.jsx
+++ b/src/shared/pages/workstack.jsx
@@ -79,15 +79,20 @@ class WorkstackPage extends Component {
         // TODO: Remove
         /* eslint-disable-next-line no-undef */
         const payload = new FormData();
-        Object.keys(formData).forEach(field => {
-            if (Array.isArray(formData[field])) {
-                formData[field].map(value => {
-                    payload.append(`${field}`, value);
+
+        for (const [key, value] of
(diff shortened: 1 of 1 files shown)


src/shared/pages/workstack.jsx
```diff
@@ -79,15 +79,20 @@ class WorkstackPage extends Component {
...
+                    payload.append(`${key}`, v);
                 });
             } else {
-                payload.append(field, formData[field]);
+                if (typeof value === 'string') {
+                    payload.append(key, value);
+                }
             }
-        });
+        }
```
robertdeniszczyc2, line 96: As this submit handler is doing the allocations, I wonder if we should dispatch a call to `updateApiStatus` in here (probably after the post) to state allocation has happened/failed. We could potentially throw it if the values aren't valid around this logic too potentially?

If we did that this could be extracted out of this handler method and specific tests written too for it, as now we're doing a bit more logic than the previous implementation.

Interested to get your thoughts on this?

robertdeniszczyc2 (approved):

PR #1247, opened 2023-02-23
HOCS-5659: Enable retrieving inactive entities in the frontend

JosephRogersonHO:


LiamMacP (approved):

server/services/form.js
```diff
@@ -134,6 +134,12 @@ async function hydrateField(field, req) {
                 choices,
                 { ...req.params, ...req.form.data }
             );
+
+            if (Object.prototype.hasOwnProperty.call(req.form, 'meta') &&
```
LiamMacP, line 138: Non-blocking:

I do wonder if there would be a better way of handling this somehow, maybe different paths for active or non-active. Maybe a good candidate for a FE review..?

JosephRogerson: Yeah, will note it down as something to look at as part of the frontend review

PR #1271, opened 2023-04-04
[HOCS-6510] Remove operational reporting list until infrastructure is in place

jeff-horton-ho-sas:
https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-6510


src/shared/reports/components/dataTable.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { AriaAttributes, useEffect, useRef } from 'react';
```
jeff-horton-ho-sas, line 1: Unused imports that were flagged as linting warnings

PR #1286, opened 2023-04-21
HOCS-4963: Remove unused min/max attributes from date

JosephRogerson:


robertdeniszczyc2:
There are also min/max year validators which might make use of the config passed in via `validator.js`, it would be worth checking to see if they can be removed aswell.

Also would be worth checking where that config is set, and whether there is any consequence of this change for cases already using min/max values.

JosephRogerson:
> There are also min/max year validators which might make use of the config passed in via `validator.js`, it would be worth checking to see if they can be removed aswell.
> 
> Also would be worth checking where that config is set, and whether there is any consequence of this change for cases already using min/max values.

The other ones are used as far as I can tell, I believe those are used when a component has isYearWithinRange as a validator which is used on a lot of dates.

The ones removed here don't actually work since the frontend update in September, because these are text elements now as specified in the design system and min and max aren't valid attributes

PR #1343, opened 2023-06-05
HOCS-6562: change document handling

LiamMacP:
Previously we're using the casework service as a permission middleware and sending the entire document through this service. This results in high memory usage and utilisation through this service, exaggerating existing performance issues.

This change adds a permission specific check that doesn't passthrough the entire file. After this permission check passes, the document is then requested from the document service directly.

### DIFF
--- a/server/routes/document.js
+++ b/server/routes/document.js
@@ -1,11 +1,11 @@
 const router = require('express').Router();
-const { getOriginalDocument, getPdfDocument, getPdfDocumentPreview } = require('../middleware/document');
+const { getDocumentInfo, getOriginalDocument, getPdfDocument, getPdfDocumentPreview } = require('../middleware/document');
 
-router.get('/:caseId/document/:documentId/preview', getPdfDocumentPreview);
+router.get('/:caseId/document/:documentId/preview', getDocumentInfo, getPdfDocumentPreview);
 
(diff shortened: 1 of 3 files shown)


server/routes/document.js
```diff
@@ -1,12 +1,12 @@
...
-router.get('/:caseId/document/:documentId/preview', getPdfDocumentPreview);
+router.get('/:caseId/document/:documentId/preview', getDocumentInfo, getPdfDocumentPreview);
 
-router.get('/:caseId/stage/:stageId/download/document/:documentId/pdf', getPdfDocument);
+router.get('/:caseId/stage/:stageId/download/document/:documentId/pdf', getDocumentInfo, getPdfDocument);
 
-router.get('/:caseId/stage/:stageId/download/document/:documentId/original', getOriginalDocument);
+router.get('/:caseId/stage/:stageId/download/document/:documentId/original', getDocumentInfo, getOriginalDocument);
 
 router.get('/:caseId/stage/:stageId/download/standard_line/:documentId', getOriginalDocument);
```
LiamMacP, line 10: This was an implicit decision to exclude from the permission check, as it doesn't contain any user information.

SWhitlam (approved):

PR #1348, opened 2023-06-06
HOCS-6656: improve date handling

LiamMacP:
Change the way that dates are handled in the process middleware. This now uses regex matching to retrieve the day, month, and year. These regexes ensure that if a non-alphanumeric character is found either side of the numbers, then the result of the regex is an empty value for the specific value. Each of the values allow for leading 0's however, these are then stripped from the result and then for day and month re-padded to length 2.

### DIFF
--- a/server/middleware/form/__tests__/process.spec.js
+++ b/server/middleware/form/__tests__/process.spec.js
@@ -368,7 +368,7 @@ describe('Process middleware', () => {
     it('should pad single digit day and months with zeros', () => {
         const req = {
             body: {
-                ['test-field']: '1989-04-03'
+                ['test-field']: '1989-4-3'
             },
             query: {},
             form: {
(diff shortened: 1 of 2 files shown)


server/middleware/form/__tests__/process.spec.js
```diff
@@ -429,10 +429,12 @@ describe('Process middleware', () => {
         expect(next).toHaveBeenCalledTimes(1);
     });
 
-    it('should not sanitise day and month with multiple connotations', () => {
```
LiamMacP, line 432: I don't believe this test is actually useful, we should just take what the user provides us and go with it.

server/middleware/form/__tests__/process.spec.js
```diff
@@ -368,7 +368,7 @@ describe('Process middleware', () => {
     it('should pad single digit day and months with zeros', () => {
         const req = {
             body: {
-                ['test-field']: '1989-04-03'
+                ['test-field']: '1989-4-3'
```
LiamMacP, line 371: This test wasn't actually testing anything before, it now is.

SWhitlam (approved):

PR #1368, opened 2023-06-30
Hocs 6656 invalid date fix

vinothon84:
HOCS-6656: Currently there is an issue when the user enters special or invalid numeric
values in date, month or year field - validation happens in javascript but it replaces 
invalid entry as blank to save it to backend DB. 

For example - if the data field entered as 2023/-05-25 
then while saving payload date is saved as -05-25 rather throwing an error in the UI. 
Also when its parsed in work stack it fails with below error:
**Cannot deserialize value of type `java.time.LocalDate` from String \"2023/-05-25\": Failed to deserialize java.time.LocalDate: (java.time.format.DateTimeParseException) Text '2023/-05-25'**

Applied fix would contain if any of the date field is incorrect, the function would return an 
empty string and this will make sure the user will get an error requesting user to enter 
valid date. Also below scenarios being handled.

**Mandatory date field**
1. If there is a special character entered in any of date field - throw error message
2. If the entered numeric is not valid then error message - for example in month fields if 13 is entered then an error is thrown
3. Only save legitimate date if entered

**Non-mandatory date field (like Complainant DOB)**
1. If a date entered just have only date field whereas month and year is blank then the field will not be saved to backend. Looks like this is the current behaviour.

### DIFF
--- a/server/middleware/form/__tests__/process.spec.js
+++ b/server/middleware/form/__tests__/process.spec.js
@@ -432,7 +432,7 @@ describe('Process middleware', () => {
     it('should return sanitised year on change', () => {
         const req = {
             body: {
-                ['test-field-1']: '0000-01-01',
+                ['test-field-1']: '2000-01-01',
                 ['test-field-2']: '00002021-01-01',
                 ['test-field-3']: '002000-01-01',
             },
@@ -475,7 +475,10 @@ describe('Process middleware', () => {
 
         processMiddleware(req, res, next);
         expect(req.form).toBeDefined();
-        expect(req.form.data).toBeUndefined();
+        expect(req.form.data).toBeDefined();
+        expect(req.form.data['test-field-1']).toEqual('2000-01-01');
+        expect(req.form.data['test-field-2']).toEqual('2021-01-01');
+        expect(req.form.data['test-field-3']).toEqual('2000-01-01');
         expect(next).toHaveBeenCalled();
         expect(next).toHaveBeenCalledTimes(1);
     });
--- a/server/middleware/form/__tests__/process.spec.js
+++ b/server/middleware/form/__tests__/process.spec.js
@@ -475,10 +475,7 @@ describe('Process middleware', () => {
 
         processMiddleware(req, res, next);
         expect(req.form).toBeDefined();
-        expect(req.form.data).toBeDefined();
-        expect(req.form.data['test-field-1']).toEqual('-01-01');
-        expect(req.form.data['test-field-2']).toEqual('2021-01-01');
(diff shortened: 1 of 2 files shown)


server/middleware/form/__tests__/process.spec.js
```diff
@@ -475,10 +475,7 @@ describe('Process middleware', () => {
 
         processMiddleware(req, res, next);
         expect(req.form).toBeDefined();
-        expect(req.form.data).toBeDefined();
-        expect(req.form.data['test-field-1']).toEqual('-01-01');
-        expect(req.form.data['test-field-2']).toEqual('2021-01-01');
-        expect(req.form.data['test-field-3']).toEqual('2000-01-01');
+        expect(req.form.data).toBeUndefined();
```
jeff-horton-ho-sas, line 478: This is undefined because next is being called with a validation error before data is set. Since validation is tested in [validation.spec.js](https://github.com/UKHomeOffice/hocs-frontend/blob/hocs-6656_invalidDateFix/server/middleware/form/__tests__/validation.spec.js) we should make sure there are tests there for this change and tests that cover `validateForm` throwing a validation error. Then if there are no valid dates that will cause this path to be followed this test can be removed. Otherwise update the dates to properly test the changed functionality.

server/middleware/form/__tests__/process.spec.js
```diff
@@ -529,10 +526,7 @@ describe('Process middleware', () => {
 
         processMiddleware(req, res, next);
         expect(req.form).toBeDefined();
-        expect(req.form.data).toBeDefined();
-        expect(req.form.data['test-field-1']).toEqual('-01-01');
-        expect(req.form.data['test-field-2']).toEqual('2021--01');
-        expect(req.form.data['test-field-3']).toEqual('2000-01-');
+        expect(req.form.data).toBeUndefined();
```
jeff-horton-ho-sas, line 529: Same as [Line 478](https://github.com/UKHomeOffice/hocs-frontend/pull/1368/files#diff-011b79a407001ece08c37cf43d3c31e1c4b91e0ab12e475f967b7ee144b78f4dR478)

jeff-horton-ho-sas (changes requested):

server/middleware/form/__tests__/process.spec.js
vinothon84: Updated with valid dates for this test.

server/middleware/form/__tests__/process.spec.js
vinothon84: Removed test case

jeff-horton-ho-sas (approved):

PR #1375, opened 2023-07-07
Spike/replace keycloak proxy

JosephRogerson:


server/middleware/auth.js
```diff
@@ -1,29 +1,34 @@
...
-                username: req.get('X-Auth-Username'),
-                id: req.get('X-Auth-UserId'),
-                groups: req.get('X-Auth-Groups'),
-                roles: req.get('X-Auth-Roles'),
-                email: req.get('X-Auth-Email'),
-                uuid: req.get('X-Auth-Subject')
-            });
+const loginCallbackMiddleware = passport.authenticate('keycloak', {
+    //TODO: is there a better way to handle this redirect?
+    successRedirect: 'back',
```
SWhitlam, line 12: It should be possible to pass the redirect URL to Keycloak and have it pass it back in the call back

PR #1389, opened 2023-07-19
HOCS:6732: New css to override divider and added div tag radio-group.jsx

vinothon84:
HOCS-6732: Introduce type divider which will be used in hocs-frontend 
to display divider. 
Respective changes added to workflow branch https://github.com/UKHomeOffice/hocs-workflow/compare/main...hocs-6732_divider_spike. 
Once dev pulls workfow & frontend for this branch, they should be able to see divider text as below screenshot.
However frontend PR needs to be merged.

image


jeff-horton-ho-sas (commented):
Looks good. I've asked Blaine to get some feedback from the requester and will approve when they give the ok.

jeff-horton-ho-sas (approved):

PR #1419, opened 2023-10-02
HOCS-6801-Bug fix to filter inactive entity lists from dropdown values

vinothon84:
HOCS-6801-Bug fix to filter inactive entity lists from dropdown values.


vinothon84:
Will replace this with another [PR](https://github.com/UKHomeOffice/hocs-frontend/pull/1425)

PR #1425, opened 2023-10-03
HOCS-6801-Bug fix to filter inactive entity lists from dropdown values

vinothon84:
HOCS-6801-Bug fix to filter inactive entity lists from dropdown values


jeff-horton-ho-sas:
Superseded by #1428

PR #1428, opened 2023-10-03
HOCS-6801:  Build changes to filter inactive entity lists. Add jest t…

vinothon84:
HOCS-6801:  This PR will include build changes to filter inactive entity lists. 
Add jest test cases to cover inactive dropdown values. This wil help not 
to display inactive entity values as part of the drop down values.

### DIFF
--- a/src/shared/common/forms/__tests__/dropdown.spec.js
+++ b/src/shared/common/forms/__tests__/dropdown.spec.js
@@ -88,23 +88,23 @@ describe('Form dropdown component', () => {
     });
 
     test('should render only active options (exclie inactive dropdown value), default is active', () => {
-        const wrapperConditionA = render(<Dropdown name="radio-group" conditionChoices={conditionChoices}
+        const wrapper = render(<Dropdown name="radio-group" conditionChoices={conditionChoices}
             data={{ 'conditionValue': 'a' }} updateState={() => null} />);
 
-        expect(wrapperConditionA).toBeDefined();
+        expect(wrapper).toBeDefined();
 
         expect(screen.getByText('conditionA valueA')).toBeInTheDocument();
         expect(screen.queryByText('conditionA valueB')).toBeInTheDocument();
         expect(screen.queryByText('conditionA valueC')).not.toBeInTheDocument();
+        expect(screen.queryByText('conditionB valueA')).not.toBeInTheDocument();
 
-        const wrapperConditionB = render(<Dropdown name="radio-group" conditionChoices={conditionChoices}
+        wrapper.rerender(<Dropdown name="radio-group" conditionChoices={conditionChoices}
             data={{ 'conditionValue': 'b' }} updateState={() => null} />);
-
-        expect(wrapperConditionB).toBeDefined();
 
         expect(screen.getByText('conditionB valueA')).toBeInTheDocument();
         expect(screen.getByText('conditionB valueB')).toBeInTheDocument();
         expect(screen.queryByText('conditionB valueC')).not.toBeInTheDocument();
+        expect(screen.queryByText('conditionA valueA')).not.toBeInTheDocument();
 
     });
 
--- a/src/shared/common/forms/__tests__/dropdown.spec.js
+++ b/src/shared/common/forms/__tests__/dropdown.spec.js
@@ -5,8 +5,8 @@ import { fireEvent, render, screen } from '@testing-library/react';
 
 const choices = [
     { label: 'isA', value: 'A' },
-    { label: 'isB', value: 'B' },
-    { label: 'isC', value: 'C' }
+    { label: 'isB', value: 'B', 'active': true },
+    { label: 'isC', value: 'C', 'active': false }
 ];
 
 // update to add active flags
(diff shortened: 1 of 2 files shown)


src/shared/common/forms/__tests__/dropdown.spec.js
```diff
@@ -9,6 +9,29 @@ const choices = [
...
+        'conditionPropertyName': 'conditionValue',
+        'conditionPropertyValue': 'b',
+        'choices': [
+            { 'label': 'isA', 'value': 'A' },
+            { 'label': 'isB', 'value': 'B', 'active': true },
+            { 'label': 'isC', 'value': 'C', 'active': false }
+        ]
+    },
+];
+
```
jeff-horton-ho-sas, line 34: Make it clearer what is being rendered in the tests by having the label include the condition value
```suggestion
const conditionChoices = [
    {
        'conditionPropertyName': 'conditionValue',
        'conditionPropertyValue': 'a',
        'choices': [
            { 'label': 'conditionA valueA', 'value': 'A' },
            { 'label': 'conditionA valueB', 'value': 'B', 'active': true },
            { 'label': 'conditionA valueC', 'value': 'C', 'active': false }
        ]
    },
    {
        'conditionPropertyName': 'conditionValue',
        'conditionPropertyValue': 'b',
        'choices': [
            { 'label': 'conditionB valueA', 'value': 'A' },
            { 'label': 'conditionB valueB', 'value': 'B', 'active': true },
            { 'label': 'conditionB valueC', 'value': 'C', 'active': false }
        ]
    },
];

```

src/shared/common/forms/__tests__/dropdown.spec.js
```diff
@@ -64,5 +87,28 @@ describe('Form dropdown component', () => {
...
+        expect(wrapper).toBeDefined();
+
+        expect(screen.getByText('isA')).toBeInTheDocument();
+        expect(screen.getByText('isB')).toBeInTheDocument();
+        expect(screen.queryByText('isC')).not.toBeInTheDocument();
+        expect(screen.queryByText('isD')).not.toBeInTheDocument();
+    });
+
+});
```
jeff-horton-ho-sas, line 114: We only need one of these tests - it should confirm that the two active from the chosen option are showing, and the inactive current choice, and the three from the other choice are not showing.

The text will need to be updated to match the label changes in the comment above.

jeff-horton-ho-sas (changes requested):
Add an additional test to confirm that options are filtered when they are not conditional. Something like

```js
test('should render only active options, default is active', () => {
    const wrapper = render(<Dropdown name="radio-group" choices={choices} updateState={() => null} />);

    expect(wrapper).toBeDefined();

    expect(screen.queryByText('isA')).not.toBeNull(); // default is active
    expect(screen.queryByText('isB')).not.toBeNull(); // explicitly active
    expect(screen.queryByText('isC')).toBeNull(); // explicitly inactive
})
```
For that to work, assigning the default choices should use filterActive too, i.e.
```js
static getChoicesToUse(defaultChoices, conditionChoices, props) {
    var choicesToUse = Dropdown.filterActive(defaultChoices);
    // ...
```
instead of
```js
static getChoicesToUse(defaultChoices, conditionChoices, props) {
    var choicesToUse = defaultChoices;
    // ...
```

src/shared/common/forms/__tests__/dropdown.spec.js
```diff
@@ -64,5 +86,36 @@ describe('Form dropdown component', () => {
...
+        expect(screen.queryByText('conditionA valueC')).not.toBeInTheDocument();
+
+        const wrapperConditionB = render(<Dropdown name="radio-group" conditionChoices={conditionChoices}
+            data={{ 'conditionValue': 'b' }} updateState={() => null} />);
+
+        expect(wrapperConditionB).toBeDefined();
+
+        expect(screen.getByText('conditionB valueA')).toBeInTheDocument();
+        expect(screen.getByText('conditionB valueB')).toBeInTheDocument();
+        expect(screen.queryByText('conditionB valueC')).not.toBeInTheDocument();
```
jeff-horton-ho-sas, line 107: We need to be testing that condition B values are not shown when condition A is selected.
```suggestion
        expect(screen.getByText('conditionB valueA')).not.toBeInTheDocument();
        expect(screen.getByText('conditionB valueB')).not.toBeInTheDocument();
        expect(screen.queryByText('conditionB valueC')).not.toBeInTheDocument();
```

I agree it's a good idea to check that the options change when the value changes. React testing library provides a way to rerender an existing component. https://testing-library.com/docs/example-update-props/

So in this case you'd need to add something like: 
```js
       // ... at end of exitsing test method, rename wrapperConditionA back to just wrapper
       wrapper.rerender(<Dropdown name="radio-group" conditionChoices={conditionChoices}
            data={{ 'conditionValue': 'b' }} updateState={() => null} />);

       expect(screen.getByText('conditionA valueA')).not.toBeInTheDocument();
       expect(screen.getByText('conditionB valueA')).toBeInTheDocument();
```

jeff-horton-ho-sas (approved):
Looks good, thanks 👍

PR #1454, opened 2023-11-23
Fix/hocs 4640 replace layout table

JosephRogerson:


JosephRogerson:
This needs merging after the axios bump as this contains that commit

vinothon84 (approved):

PR #1498, opened 2024-03-25
Label added - HTML checker passed

RobertHerringHO:
https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-4967

Updates:

- Label attribute added to dropdown return options.
- Ternary operator added to avoid bypass of required dropdown selections.
- Snapshot updated to allow test pass.

### DIFF
--- a/src/shared/common/forms/dropdown.jsx
+++ b/src/shared/common/forms/dropdown.jsx
@@ -127,8 +127,8 @@ class Dropdown extends Component {
                         return (
                             <option
                                 key={i}
-                                value={choice.value === '' ? null : choice.value}
-                                label={choice.label}
+                                value={choice.value}
+                                label={choice.label || 'No option selected'}
                             >
                                 {choice.label}
                             </option>
--- a/src/shared/common/forms/dropdown.jsx
+++ b/src/shared/common/forms/dropdown.jsx
@@ -125,7 +125,13 @@ class Dropdown extends Component {
                 >
                     {choicesToUse && choicesToUse.map((choice, i) => {
                         return (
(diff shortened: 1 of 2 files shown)


src/shared/common/forms/dropdown.jsx
```diff
@@ -125,7 +125,13 @@ class Dropdown extends Component {
                 >
                     {choicesToUse && choicesToUse.map((choice, i) => {
                         return (
-                            <option key={i} value={choice.value} >{choice.label}</option>
+                            <option
+                                key={i}
+                                value={choice.value === '' ? null : choice.value}
+                                label={choice.label}
```
jeff-horton-ho-sas, line 131: The value should be left as-is to preserve behavior, the label here is what a screen reader will read out for that option. 

To include a label attribute on every option this would work:

```suggestion
                                value={choice.value}
                                label={choice.label || 'No option selected'}
```

A screen reader will use the contents of the option if present, which makes including a label that matches the contentes redundant. We only need to include a label attribute when the option will be empty. This is because whilst the select being empty visually indicates that it is empty, the screen reader just doesn't say anything, so the user doesn't get told about it, so we need to add a non-empty label explaining there is an empty option.  Passing null tells React to omit that attribute, so we can do that except when there is no label like so:

```suggestion
                                value={choice.value}
                                label={choice.label ? null : 'No option selected' }
```
Go with the verision seems clearest to you.

jeff-horton-ho-sas (changes requested):

jeff-horton-ho-sas (approved):

PR #1500, opened 2024-04-08
HOCS-4908: Render tabs on small screen

JosephRogerson:


src/shared/common/components/side-bar.jsx
```diff
@@ -129,7 +174,8 @@ class SideBar extends Component {
                             })
                         }
                     </ul>
-                    {this.renderActiveTab()}
+                    {this.state.renderAll && this.renderAllTabs()}
+                    {!this.state.renderAll && this.renderActiveTab()}
```
jeff-horton-ho-sas, line 178: I think it would be clearer that this is an either/or choice with a ternary operator
```suggestion
                    {this.state.renderAll ? this.renderAllTabs() : this.renderActiveTab()}
```

jeff-horton-ho-sas (approved):
Looks good, one minor comment. Thanks.

